PR #91, opened 2023-02-23
Header variables to scss alias

rchavan10:


axdanbol:
How to theme the header needs some more thought put into it so I'm closing this pr in the meanwhile

PR #93, opened 2023-02-24
Email input

hsjoshi28:


axdanbol:
@hsjoshi28 You are doing way to much for this issue. You should only be doing the email input box, not any other parts of the help modal

PR #112, opened 2023-03-03
hra-download-state behavioral

sanket1211996:


libs/state/src/lib/download-file/download-file.state.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Actions download file in specified format
+   * @param ctx
+   * @param action
+   */
+  @Action(DownloadFile)
+  async downloadFile(ctx: StateContext<DownloadStateModel>, action: DownloadFile) {
+    const fileName = fileUrl.substring(fileUrl.lastIndexOf('/') + 1, fileUrl.lastIndexOf('.'));
+
```
axdanbol, line 48: If url is available then download directly. Otherwise if format is supported convert

axdanbol (changes requested):

sanket1211996:
@bherr2  @axdanbol Code is ready for review there is some issue with checks. Could you please check.

libs/state/src/lib/download/download.model.ts
```diff
@@ -0,0 +1,26 @@
...
+   */
+  format: FileFormat;
+
+  /**
+   * label display in UI
+   */
+  label: string;
+}
+
+export type DownloadModel = DownloadFormat[];
```
axdanbol, line 26: Add doc comment

libs/state/src/lib/download/index.ts
```diff
@@ -0,0 +1,3 @@
+export * as DownloadAction from './download.action';
+export { DownloadState } from './download.state';
+export { FileFormat, DownloadFormat } from './download.model';
```
axdanbol, line 3: Export selectors

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,149 @@
...
+    },
+  ],
+})
+export class DownloadState {
+  /**
+   * Download converted formatted file to browser
+   * @param blob
+   * @param fileName
+   */
+  blobDownload(blob: Blob, fileName: string) {
```
axdanbol, line 32: Rename to `downloadData`
Should also be private

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,149 @@
...
+  inferFilename(fileUrl: string) {
+    return fileUrl.substring(fileUrl.lastIndexOf('/') + 1, fileUrl.lastIndexOf('.'));
+  }
+
+  /**
+   * Infers file extension from file url
+   * @param fileUrl
+   * @returns
+   */
+  inferFileExtension(fileUrl: string) {
```
axdanbol, line 57: Don't need to guess extension. You get that from the format

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,149 @@
...
+    return fileUrl.substring(fileUrl.lastIndexOf('.') + 1);
+  }
+
+  /**
+   * Downloads and save -  method is used to direct fetch file
+   * if available on url without conversion
+   * @param fileUrl
+   * @param fileName
+   */
+  async downloadAndSave(fileUrl: string, fileName: string) {
```
axdanbol, line 67: Rename to `downloadRemoteData`

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,149 @@
...
+    anchor.remove();
+    window.URL.revokeObjectURL(url);
+  }
+
+  /**
+   * Infers filename from file url
+   * @param fileUrl
+   * @returns
+   */
+  inferFilename(fileUrl: string) {
```
axdanbol, line 48: Rename to `guessFilename`
Also it doesn't work as intended. Try the following examples:
`abc`
`abc/def`
`abc/def/`

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,149 @@
...
+   * Creates canvas of svg from HTMLImageElement
+   * @param surl
+   * @returns canvas
+   */
+  async createCanvas(image: HTMLImageElement): Promise<HTMLCanvasElement> {
+    const canvas = document.createElement('canvas');
+    const ctx = canvas.getContext('2d');
+    canvas.width = image.width;
+    canvas.height = image.height;
+    ctx?.drawImage(image, 0, 0);
```
axdanbol, line 122: Should fail if we can't draw it (i.e. no context)

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,149 @@
...
+    await image.decode();
+    return image;
+  }
+
+  /**
+   * Creates canvas of svg from HTMLImageElement
+   * @param surl
+   * @returns canvas
+   */
+  async createCanvas(image: HTMLImageElement): Promise<HTMLCanvasElement> {
```
axdanbol, line 117: Rename to `convertImageToCanvas`

axdanbol (changes requested):

libs/state/src/lib/download/download.action.ts
```diff
@@ -0,0 +1,44 @@
...
+export class AddEntry extends Action('Add') {
+  constructor(readonly id: DownloadFormatId, readonly entry: DownloadEntry) {
+    super();
+  }
+}
+
+/**
+ * Clear entry from download state
+ */
+export class ClearEntry extends Action('Clear') {
```
axdanbol, line 27: I rather have an action that clears all entries (i.e. sets entries to the empty object {}) than having to clear each one by one

libs/state/src/lib/download/download.selectors.ts
```diff
@@ -0,0 +1,20 @@
+import { Selector } from '@ngxs/store';
+import { DownloadFormat, DownloadModel } from './download.model';
+
+/**
+ * Available format selectors
+ */
+export class DownloadSelectors {
+  /**
+   * Selectors available format for download selectors
+   * @param state
```
axdanbol, line 10: Add docs to param and return tags

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,144 @@
...
+ */
+@State<DownloadModel>({
+  name: 'download',
+  defaults: {
+    formats: {},
+    entries: {},
+  },
+})
+export class DownloadState implements NgxsOnInit {
+  private readonly http = inject(HttpClient);
```
axdanbol, line 24: Code doc

libs/state/src/lib/download/download.state.ts
```diff
@@ -0,0 +1,144 @@
...
+    const { entries } = ctx.getState();
+    const entry = entries[format];
+    let filename = `download${format}`;
+    switch (entry?.type) {
+      case 'url':
+        filename = this.guessFilename(ctx, format, entry.url);
+        return this.downloadRemoteData(entry.url).pipe(tap((data) => this.downloadData(data, filename)));
+
+      case 'data':
+        this.downloadData(new Blob([entry.data]), filename);
```
axdanbol, line 92: Use guessFilename with an empty url '' for the filename here

axdanbol (changes requested):

axdanbol (approved):

PR #113, opened 2023-03-06
Override mat button

rchavan10:


axdanbol (changes requested):
- Button still looks wonky on active
  - [ ] There's a darker rectangle in the middle
  - [ ] The outline is missing the white piece
- Anchor
  - [ ] Shouldn't have background on hover
  - [ ] Doesn't change when disabled
  - [ ] Wonky looking on active
  - [ ] Add attribute `tabindex="1"` in story to more easily focus link

libs/shared/theming/src/lib/styles/_overrides.scss
```diff
@@ -0,0 +1,85 @@
+@use 'variables' as vars;
+
+@mixin init() {
+  @at-root {
+    button.mdc-button {
```
axdanbol, line 5: Don't forget to convert px to rem. Also add newlines between selectors

libs/shared/theming/src/lib/styles/_overrides.scss
```diff
@@ -0,0 +1,121 @@
...
+        background-color: #e7e7e5;
+        border: 0.063rem #9d9e9a solid;
+        color: #9d9e9a;
+        :hover {
+          cursor: not-allowed;
+        }
+      }
+
+      .mat-mdc-button-persistent-ripple::before {
+        opacity: 0;
```
axdanbol, line 79: Add important

axdanbol (approved):

PR #116, opened 2023-03-07
Resource state

jsaikiran19:


libs/cdk/states/src/resource-registry/resource-registry.model.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+/** type to hold Resource object for markdown */
+export interface MarkdownResource {
+  /** indicates type for resource */
+  type: ResourceType.Markdown;
+  /** markdown content */
+  markdown: string;
+}
+
```
axdanbol, line 18: Add interface for UrlResource

libs/cdk/states/src/resource-registry/resource-registry.state.ts
```diff
@@ -0,0 +1,62 @@
...
+  loadMarkdown(ctx: StateContext<ResourceRegistryModel>, { id, url }: LoadMarkdown): void {
+    this.http
+      .get(url, { responseType: 'text' })
+      .pipe(
+        tap((data) => {
+          const entry = { type: ResourceType.Markdown, markdown: data };
+          this.addResourceEntry(ctx, id, entry);
+        })
+      )
+      .subscribe();
```
axdanbol, line 45: Return the observable instead of subscribing. Ngxs subscribes to returned values automatically

axdanbol (approved):

PR #129, opened 2023-03-16
adds Link registry

jsaikiran19:


axdanbol (changes requested):
Let's have a look at internal navigation in storybook together when you're in the office

axdanbol (approved):

PR #132, opened 2023-03-16
google-analytics

Hanuhitha:
Merge develop to main


libs/state/src/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './lib/contact/contact.actions';
 export * from './lib/contact/contact.state';
 export * as ResourceIds from './lib/resource-ids';
 export * from './lib/source-list/source-list.state';
+export { StateAnalyticsPluginService } from './lib/state-analytics-plugin/state-analytics-plugin.service';
```
axdanbol, line 5: Also needs to export the provider token

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class StateAnalyticsPluginService implements NgxsPlugin {
+  /**
+   * Creates an instance of logger plugin.
+   * @param ga
+   */
+  constructor(private readonly ga: GoogleAnalyticsService) {}
```
axdanbol, line 25: Use `inject` instead

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Handles logger plugin
+   * @param state
+   * @param action
+   * @param next
+   * @returns
+   */
+  handle(state: unknown, action: unknown, next: NgxsNextPluginFn) {
+    console.log('In handle');
```
axdanbol, line 35: Remove logs

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Logs action trigerred to google analytics event
+   * @param action
+   * @returns action
+   */
+  private logAction(action: unknown): void {
+    const type = getActionTypeFromInstance(action);
+    console.log('type', type);
+
+    if (type) {
```
axdanbol, line 49: We should only log approved actions. Not all of them

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * Serializes action data
+   * @param this
+   * @param key
+   * @param value
+   * @returns serialize
+   */
+  private serialize(this: void, key: unknown, value: unknown): unknown {
```
axdanbol, line 63: @bherr2 Should we update the serialization logic to include nested objects, etc.?

axdanbol (changes requested):

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
axdanbol: Not resolved. Type needs to be checked against approved action list/set to determine if the action should be logged

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
axdanbol: Not resolved!

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
bherr2: First argument to ga.event (See <https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#ec>) has a 150 byte limit, second has a 500 byte limit, 3rd also has a 500 byte limit). If it goes over, it gets truncated.

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
```diff
@@ -0,0 +1,99 @@
...
+});
+
+/**
+ * Google Analytics Plugin Injectable
+ */
+@Injectable({
+  providedIn: 'root',
+})
+
+/**
```
axdanbol, line 33: Remove this comment. Class comment appears before decorator

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Logs action trigerred to google analytics event
+   * @param action
+   * @returns action
+   */
+  private logAction(action: unknown): void {
+    const type = getActionTypeFromInstance(action);
+    if (type && this.loggableTypes.has(type)) {
+      console.log('Received Action', action);
```
axdanbol, line 68: Remove log

axdanbol (changes requested):
Increase test coverage to pass checks

libs/state/src/lib/state-analytics-plugin/state-analytics-plugin.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    return undefined;
+  }
+}
+
+/**
+ * State Login Provider
+ */
+export const STATE_LOGGER_PROVIDER: Provider = {
+  provide: NGXS_PLUGINS,
+  useClass: StateAnalyticsPluginService,
```
axdanbol, line 91: Update this property to `useExisting` instead

axdanbol (changes requested):

axdanbol (approved):

PR #136, opened 2023-03-23
Hover Directive

pranathi-neerudu:


libs/cdk/src/lib/hover/hover.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+  standalone: true,
+})
+export class HoverDirective<T = unknown> {
+  /** Input for the template portal */
+  @Input('hraHover')
+  set content(content: TemplateRef<HoverContext<T>>) {
+    this.portal = new TemplatePortal(content, this.viewContainerRef);
+    this.updateContent();
+  }
+
```
axdanbol, line 25: Add an input called `hraHoverData` and update the template's data when it changes

axdanbol (approved):

axdanbol (changes requested):
Need to improve documentation comments. They should all read like complete sentences that clearly explain what the item is for without repeating information already available by typescript.
Ex. Instead of 
```ts
/**  Interface for the hover context */
export interface HoverContext<T = unknown> {
  /** defining an implicit type */
  $implicit: T;
}
```
do
```ts
/** Context passed to hover content templates */
export interface HoverContext<T = unknown> {
  /** Data provided by the user for use inside the content template */
  $implicit: T;
}

axdanbol (approved):

PR #160, opened 2023-04-25
tissue-service

jsaikiran19:


libs/services/src/lib/tissue-ftu/tissue-ftu.service.ts
```diff
@@ -0,0 +1,17 @@
+import { HttpClient } from '@angular/common/http';
+import { Injectable, inject } from '@angular/core';
+import { Observable } from 'rxjs';
+
+/** Service to fetch reference organs */
+@Injectable({
+  providedIn: 'root',
+})
+export class TissueFtuService {
```
axdanbol, line 9: Make base service abstract and create a mock implementation service

libs/state/src/lib/medical-illustration/medical-illustration.model.ts
```diff
@@ -19,8 +19,18 @@ const mapEntry = z.object({
   'Inset #': z.string(),
 });
 
+/** Reference organ type */
+const referenceOrgan = z.object({
```
axdanbol, line 23: Move to service file so you can use the type in getReferenceOrgans()

libs/state/src/lib/medical-illustration/medical-illustration.state.ts
```diff
@@ -31,6 +31,24 @@ export class MedicalIllustrationState {
...
 
+  /**
+   * Sets uri by finding it from reference organs
+   * @param param0 state context
+   * @param param1 Action object with iri
+   */
+  @Action(SetUriFromIRI)
+  setUriFromIRI({ patchState, getState }: MedicalIllustrationContext, { iri }: SetUriFromIRI) {
+    const referenceOrgans = getState().referenceOrgans;
+    if (referenceOrgans?.length) {
```
axdanbol, line 42: Use `referenceOrgans?.find(...)` and throw an error if no matching organ is found

axdanbol (changes requested):

PR #163, opened 2023-04-26
Hra tissue library state

rchavan10:


libs/state/tsconfig.spec.json
```diff
@@ -7,5 +7,11 @@
     "target": "es2016"
   },
   "files": ["src/test-setup.ts"],
-  "include": ["jest.config.ts", "src/**/*.test.ts", "src/**/*.spec.ts", "src/**/*.d.ts"]
+  "include": [
+    "jest.config.ts",
+    "src/**/*.test.ts",
+    "src/**/*.spec.ts",
+    "src/**/*.d.ts",
+    "../services/src/lib/tissue-library/tissue-library.service.spec.ts"
```
axdanbol, line 15: Why are you adding this?

axdanbol (changes requested):

libs/state/tsconfig.spec.json
rchavan10: I am not sure how it went there, I will remove it.

axdanbol (dismissed):

axdanbol (approved):

PR #188, opened 2023-06-19
added source state to illustration

rchavan10:


axdanbol:
Seems to have been fixed during state refactoring

PR #209, opened 2023-08-01
added box-shadow

rchavan10:


libs/components/molecules/src/lib/landing-page-in-depth/landing-page-in-depth.component.scss
```diff
@@ -14,6 +14,8 @@
     width: 100%;
     height: auto;
     align-self: center;
+    border-radius: 16px;
```
axdanbol, line 17: Please change all px values into rem

axdanbol (changes requested):

libs/components/molecules/src/lib/landing-page-in-depth/landing-page-in-depth.component.scss
rchavan10: Will do that now

PR #211, opened 2023-08-03
Medical illustration webcomponent

edlu77:


apps/medical-illustration/src/assets/TEMP/illustration-data.json
bherr2, line 1: Please rename to `2d-ftu-illustrations.jsonld`

apps/medical-illustration/src/app/app-web-component.ts
```diff
@@ -0,0 +1,103 @@
...
+})
+export class AppWebComponent implements OnChanges {
+  /** Http client */
+  private readonly http = inject(HttpClient);
+
+  /** Ftu illustrations file: can be url or file object */
+  @Input() lookupSrc: string | JsonLd = '';
+
+  /** Illustration info: can be illustration id or illustration data object */
+  @Input() illustrationSrc: string | IllustrationData = '';
```
bherr2, line 25: Once fully published, we will set a default illustration source with a stable URL. Will that work here?

apps/medical-illustration/webcomponent-example.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Medical Illustration Web Component</title>
+    <base href="/" />
```
bherr2, line 6: Does this web component rely on base href? It really shouldn't. Needs to be able to work flexibly and standalone. I think it's true here, but just want to check.

edlu77: I don't think that needs to be there, will remove it.

apps/medical-illustration/src/app/app-web-component.ts
edlu77: By illustration source you mean the entire ftu illustrations file right? In that case you set the url as lookupSrc and it will fetch the data.

bherr2: Yes, that's right. When we get a stable URL for the jsonld file we will put that there.

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.ts
```diff
@@ -106,6 +116,48 @@ export class InteractiveSvgComponent<T extends NodeMapEntry> implements OnDestro
...
+   * @param changes
+   */
+  ngOnChanges(changes: SimpleChanges): void {
+    if ('highlightId' in changes && this.crosswalkEl) {
+      this.resetHighlight();
+      const cellMatch = this.mapping?.find((entry) => entry.id === this.highlightId); // find the matching cell data in mapping
+      let crosswalkMatch: NodeListOf<Element>;
+      if (cellMatch) {
+        const name = cellMatch.name;
+        const childId = name.replace(/_/g, '_x5F_'); // get the id of the matching cell in svg
```
axdanbol, line 133: Needs to do the reverse of decodeId

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.ts
```diff
@@ -106,6 +116,48 @@ export class InteractiveSvgComponent<T extends NodeMapEntry> implements OnDestro
...
+        if (elementMatch) {
+          if (elementMatch.nodeName === 'g') {
+            // highlights children if g element
+            crosswalkMatch = this.crosswalkEl.querySelectorAll(`#${childId} :is(path, polygon, polyline)`);
+          } else {
+            // highlights siblings if path, polygon, or polyline element
+            const parentId = elementMatch.parentElement?.id ?? '';
+            crosswalkMatch = this.crosswalkEl.querySelectorAll(`#${parentId} :is(path, polygon, polyline)`);
+          }
+          crosswalkMatch.forEach((element) =>
```
axdanbol, line 144: It would be better to add a class to each element rather than setting styles. That way the styles can be easily changed in the scss file. Then you would remove the class on resetHightlight

apps/medical-illustration/src/app/app-web-component.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Gets illustration data and sets illustration url and mapping data
+   * @param illustrationSrc Illustration src
+   * @param lookupSrc Lookup src
+   * @returns Observable
+   */
+  private getData(illustrationSrc: IllustrationData | string, lookupSrc: string): Observable<JsonLd> {
+    return this.http.get<JsonLd>(lookupSrc).pipe(
```
axdanbol, line 65: Don't you have to specify that the `.get(...)` should return json or is that the default?

apps/medical-illustration/src/app/app-web-component.ts
```diff
@@ -0,0 +1,119 @@
...
+  @Input() highlightId = '';
+
+  /** Emits node data when node hovered */
+  @Output() readonly nodeHovered = new EventEmitter<NodeMapEntry>();
+
+  /** Emits node data when node clicked */
+  @Output() readonly nodeClicked = new EventEmitter<NodeMapEntry>();
+
+  /** Illustration svg url behavior subject */
+  readonly url$ = new BehaviorSubject<string>('');
```
axdanbol, line 37: The way you are using observables is unnecessary and can be replaced by regular properties

apps/medical-illustration/src/app/app-web-component.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Finds and sets illustration url and mapping data
+   * @param illustrationFile Illustration data file
+   * @param illustrationSrc Illustration src
+   */
+  private setUrlAndMapping(illustrationFile: JsonLd, illustrationSrc: IllustrationData | string) {
+    if (typeof illustrationSrc === 'string') {
+      // input src is id
+      const currentOrganData = illustrationFile['@graph'].find((entry) => entry['@id'] === this.illustrationSrc);
```
axdanbol, line 80: Some of this code may be able to take advantage of the ftu data service

axdanbol (changes requested):
Also merge and resolve conflicts with develop

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.html
```diff
@@ -16,7 +16,17 @@
     [cdkConnectedOverlayViewportMargin]="16"
     [cdkConnectedOverlayOpen]="true"
   >
-    <div class="tooltip">
+    <div
+      class="tooltip"
+      style="
```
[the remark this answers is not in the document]
edlu77: It doesn't work when I do that, this was the only way I could get the tooltip styled 🤷‍♂️

apps/medical-illustration/src/app/app-web-component.ts
axdanbol: Replace url$ with a regular property like you did with mapping

apps/medical-illustration/src/app/app-web-component.ts
```diff
@@ -0,0 +1,114 @@
...
+      if (currentOrganData) {
```
axdanbol, line 77: This block and the block in the else case are virtually the same. Move to a method to reduce duplication

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.html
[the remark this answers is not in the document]
axdanbol: Then make a new atom component and move styles into it, then use that component instead

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.ts
```diff
@@ -204,6 +262,16 @@ export class InteractiveSvgComponent<T extends NodeMapEntry> implements OnDestro
     return id.replace(/_x([\da-f]+)_/gi, replacer);
   }
 
+  /**
+   * Turns normal string into decoded SVG id
+   * @param id id
+   * @returns Decoded id
+   */
+  private unDecodeId(id: string): string {
```
axdanbol, line 270: Usually called 'encode'...

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.ts
```diff
@@ -204,6 +262,16 @@ export class InteractiveSvgComponent<T extends NodeMapEntry> implements OnDestro
...
+    const replacer = (_match: string, pos: number) => `_x${id.charCodeAt(pos).toString(16).toUpperCase()}_`;
```
axdanbol, line 271: Use the match argument instead of id + pos, i.e. `match.charCodeAt(0)`

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.ts
```diff
@@ -204,6 +262,16 @@ export class InteractiveSvgComponent<T extends NodeMapEntry> implements OnDestro
...
+    return id.replace(/_/g, replacer);
```
axdanbol, line 272: Use `/[^a-z0-9-]/gi` instead to replace all non alphanumeric/hyphens characters

libs/components/molecules/src/lib/interactive-svg/interactive-svg.component.ts
```diff
@@ -106,6 +116,53 @@ export class InteractiveSvgComponent<T extends NodeMapEntry> implements OnDestro
...
+    if ('highlightId' in changes) {
+      this.resetHighlight();
+      this.setHighlight();
+    }
+  }
+
+  /**
+   * Highlights cells that match highlightId
+   */
+  private setHighlight() {
```
axdanbol, line 136: This method needs some refactoring/cleanup to reduce nesting/duplication. @edlu77 Ping me on slack and we can meet to go over it

axdanbol (changes requested):
More changes

PR #223, opened 2023-08-16
Data service implementation

Knight7561:


libs/services/src/lib/ftu-data/ftu-data.impl.ts
```diff
@@ -0,0 +1,371 @@
...
+/** Constant  to read the endpoints */
+export const FTU_DATA_IMPL_ENDPOINTS = new InjectionToken<FtuDataImplEndpoints>('Endpoints');
+
+/** Input to different file formats supported */
+export const FTU_DATA_IMPL_FILE_FORMAT_MAPPING = new InjectionToken<Record<string, string>>('Mapping of file formats', {
+  providedIn: 'root',
+  factory: () => ({
+    'image/svg+xml': 'svg',
+    'image/png': 'png',
+    'application/pdf': 'pdf',
```
bherr2, line 34: I know this is weird, but application/pdf in this context is actually an adobe illustrator file (ai)

Knight7561: Okay sure. WIll refactor that

axdanbol (approved):

PR #230, opened 2023-08-28
example web components

jsaikiran19:


bherr2:
Do these work?

bherr2:
> > Do these work?
> 
> These are just example htmls bruce. It won't load the application. It's just for reference on how to use the web component html.

If they don't work, they are _terrible_ examples.

jsaikiran19:
> > > Do these work?
> > 
> > 
> > These are just example htmls bruce. It won't load the application. It's just for reference on how to use the web component html.
> 
> If they don't work, they are _terrible_ examples.

Yeah that was the initial draft for those htmls. I don't think I've updated them. I just updated the ftu-large and ftu-small htmls with the latest changes.

bherr2 (approved):

PR #280, opened 2023-09-19
notice for window resize

rchavan10:


bherr2:
This does not work right. Need someone to pick up getting this done.

Issue #169, opened 2023-05-08
Tissue Library - Improve navigation when using a keyboard


axdanbol:
Most of this should probably be possible to accomplish by correcting aria attributes
- The up/down keys should move between selectable items
- The left/right keys should expand/collapse the current node if it has children

PR #296, opened 2023-10-25
adds navigation using up/down arrows

jsaikiran19:
#169

### DIFF
--- a/libs/components/behavioral/src/lib/tissue-library-behavior/tissue-library-behavior.component.ts
+++ b/libs/components/behavioral/src/lib/tissue-library-behavior/tissue-library-behavior.component.ts
@@ -1,10 +1,11 @@
 import { CommonModule } from '@angular/common';
-import { ChangeDetectionStrategy, Component, ViewChild } from '@angular/core';
-import { select$, selectSnapshot } from '@hra-ui/cdk/injectors';
+import { ChangeDetectionStrategy, Component, HostListener, ViewChild } from '@angular/core';
+import { dispatch, select$, selectSnapshot } from '@hra-ui/cdk/injectors';
 import { TissueTreeListComponent } from '@hra-ui/components/molecules';
 import { Tissue } from '@hra-ui/services';
 import { ActiveFtuSelectors, TissueLibrarySelectors } from '@hra-ui/state';
 import { LabelBoxComponent } from '@hra-ui/components/atoms';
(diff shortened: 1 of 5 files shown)


libs/components/behavioral/src/lib/tissue-library-behavior/tissue-library-behavior.component.ts
```diff
@@ -48,4 +51,50 @@ export class TissueLibraryBehaviorComponent {
       }
     });
   }
+
+  @HostListener('document:keydown', ['$event'])
```
axdanbol, line 55: Should only happen when this component has focus.

libs/components/behavioral/src/lib/tissue-library-behavior/tissue-library-behavior.component.ts
```diff
@@ -48,4 +51,50 @@ export class TissueLibraryBehaviorComponent {
...
+      if (event.key === 'ArrowDown' && selectedIndex + 1 < nodes.length) {
+        this.navigateToNode(nodes[selectedIndex + 1]);
+      }
+      if (event.key === 'ArrowUp' && selectedIndex - 1 >= 0) {
+        this.navigateToNode(nodes[selectedIndex - 1]);
+      }
+    }
+  }
+
+  navigateToNode(node: any): void {
```
axdanbol, line 94: node needs better typing

libs/components/behavioral/src/lib/tissue-library-behavior/tissue-library-behavior.component.ts
```diff
@@ -48,4 +51,50 @@ export class TissueLibraryBehaviorComponent {
       }
     });
   }
+
```
axdanbol, line 54: Move all this logic into the pure component and add an output. Might even be able to programmatically emit a click on a node to trigger navigation through the hraLink

libs/components/behavioral/src/lib/tissue-library-behavior/tissue-library-behavior.component.ts
axdanbol, line 94: Should only navigate when the user is on the illustration page

axdanbol (changes requested):
Highlighting and keyboard navigation should preferable also work on the homepage

PR #357, opened 2024-04-26
create vis app header

bhushankhope:


bhushankhope:
Added color picker (work in progress)

libs/cde-visualization/package.json
```diff
@@ -2,11 +2,12 @@
   "name": "@hra-ui/cde-visualization",
   "version": "0.0.1",
   "peerDependencies": {
+    "@angular/cdk": "17.3.4",
     "@angular/common": "^17.3.0",
-    "@angular/core": "^17.3.0"
+    "@angular/core": "^17.3.0",
```
axdanbol, line 7: Update the core and common version too

libs/cde-visualization/src/index.ts
```diff
@@ -1 +1,17 @@
+import { createApplication } from '@angular/platform-browser';
+import { createCustomElement } from '@angular/elements';
+import { CdeVisualizationComponent } from './lib/cde-visualization/cde-visualization.component';
+
+(async () => {
```
axdanbol, line 5: Annoyingly @angular/elements have not been updated to work with signal based components yet... I'm currently working on a utility library that will make this simpler and work around the bug. Meanwhile continue to use the component as a regular angular component instead of trying to use it as a web component.

libs/cde-visualization/src/lib/cde-visualization/cde-visualization.component.ts
```diff
@@ -1,12 +1,182 @@
...
   standalone: true,
-  imports: [CommonModule],
+  imports: [CommonModule, VisualizationHeaderComponent, MetadataComponent, CellTypesComponent],
   templateUrl: './cde-visualization.component.html',
   styleUrl: './cde-visualization.component.scss',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class CdeVisualizationComponent {}
+export class CdeVisualizationComponent {
+  readonly nodes = input<Node[]>([
```
axdanbol, line 32: We have to update this to work both with an array input or a file url input. It is probably easiest if we update this together over a zoom call.

libs/cde-visualization/src/lib/cde-visualization/cde-visualization.component.ts
```diff
@@ -1,12 +1,182 @@
...
+    {
+      x: 6789.012,
+      y: 3456.789,
+      cell_type: 'Osteocyte',
+    },
+  ]);
+
+  readonly cellTypeAnchor = input<string>();
+
+  readonly colorMap = input<ColorMapItem[]>();
```
axdanbol, line 137: Like the `nodes` input this need to accept a url as well

axdanbol (changes requested):

libs/cde-visualization/package.json
bhushankhope: updated versions

libs/cde-visualization/src/lib/cde-visualization/cde-visualization.component.ts
bhushankhope: now works with both string and nodes[]

libs/cde-visualization/src/lib/cde-visualization/cde-visualization.component.ts
bhushankhope: now accepts url

axdanbol (approved):

Issue #344, opened 2024-04-05
CDE: Vis App


LibbyUX:
# CDE MVP1: Vis App

## User Stories
- As a pediatric pathologist, I want single-cell data, in order to improve disease diagnosis.

## Resources
- Design File: [Vis App 1920px](https://www.figma.com/file/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?type=design&node-id=584-12139&mode=design&t=bgmiG3vgdAwXQihE-4)
- Prototype Presentation: [Vis App](https://www.figma.com/proto/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?type=design&node-id=1-25&viewport=-2119%2C2133%2C0.42&t=bgmiG3vgdAwXQihE-0&scaling=scale-down&starting-point-node-id=584%3A12139&show-proto-sidebar=1)
- HRA Design System: [Component Library](https://www.figma.com/file/gQEMLugLjweDvbsNNUVffD/Human-Reference-Atlas-Design-System?type=design&node-id=103-4&mode=design)
- HRA Design System: [Style Guide](https://www.figma.com/file/gQEMLugLjweDvbsNNUVffD/Human-Reference-Atlas-Design-System?type=design&node-id=0-1&mode=design)

## Main Components
- [CDE App header](https://www.figma.com/file/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?type=design&node-id=550-11480&mode=design&t=bgmiG3vgdAwXQihE-4)
- [Tooltip hover cards](https://www.figma.com/file/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?type=design&node-id=605-26344&mode=design&t=bgmiG3vgdAwXQihE-4)
- [Metadata Information module](https://www.figma.com/file/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?type=design&node-id=584-12575&mode=design&t=bgmiG3vgdAwXQihE-4)
- [Cell Types module](https://www.figma.com/file/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?type=design&node-id=811-20240&mode=design&t=bgmiG3vgdAwXQihE-4)
[...]

PR #370, opened 2024-05-03
Vis app histogram

edlu77:
Histogram component of visualization app (#344)


bherr2:
Any reason why you're not using the built-in export to SVG and PNG that vega-lite/vega-embed provides?

libs/cde-visualization/src/lib/histogram/histogram.component.ts
```diff
@@ -0,0 +1,207 @@
...
+          legend: null,
+          scale: { range: this.colorMap.map((entry) => this.rgbToHex(entry.cell_color)) },
+        },
+      },
+    };
+  }
+
+  download(event: MouseEvent, type: 'svg' | 'png') {
+    event.stopPropagation();
+    embed(this.vis?.nativeElement, this.spec).then((result) => {
```
bherr2, line 194: You shouldn't have to recreate the visualization here. Save the view in line 145 and use it here instead of using embed again.

apps/cde-ui/src/assets/color_mapping.csv
bherr2, line 1: These assets should go in assets/TEMP/ as they will be removed once the application is fully functional

PR #449, opened 2024-06-14
Add diversity survey to release notes

andreasbueckle:


bherr2:
That doesn't fix it. We'll have a dev do this in another branch.

PR #485, opened 2024-06-24
fix netlify routing

bhushankhope:


apps/humanatlas.io/project.json
```diff
@@ -23,10 +23,16 @@
         "assets": [
           "apps/humanatlas.io/src/favicon.png",
           "apps/humanatlas.io/src/assets",
+          "apps/humanatlas.io/src/_redirects",
```
bherr2, line 26: Revert changes in this file (what you're doing in github workflows is the better solution)

apps/humanatlas.io/src/_redirects
```diff
@@ -0,0 +1 @@
+/*  /index.html 200
```
bherr2, line 1: Remove this file (what you're doing in github workflows is the better solution)

bherr2 (changes requested):

.github/workflows/build-previews.yml
```diff
@@ -18,6 +18,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Make create_redirects.sh executable
+        run: chmod +x ./.github/workflows/create-redirects.sh
```
bherr2, line 22: You shouldn't need to do this here, just mark the sh file as executable on command line locally and commit the change:
```
chmod +x ./.github/workflows/create-redirects.sh
git commit -a
```

apps/humanatlas.io/project.json
```diff
@@ -27,6 +27,11 @@
             "glob": "**/*",
             "input": "apps/humanatlas.io/src/static",
             "output": "."
+          },
```
bherr2, line 30: Remove

bherr2 (approved):

PR #495, opened 2024-06-26
add base href to dashboard-ui

bhushankhope:
Added base href to dashboard

### DIFF
--- a/apps/dashboard-ui/src/app/app.component.ts
+++ b/apps/dashboard-ui/src/app/app.component.ts
@@ -1,7 +1,8 @@
-import { Component } from '@angular/core';
+import { Component, OnInit, input } from '@angular/core';
 import { RouterModule } from '@angular/router';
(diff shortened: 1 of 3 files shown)


bherr2:
Let's wait for @axdanbol to get back on Monday. I'm sure he'll have a great solution for this and probably best to wait until he can weigh in.

bherr2:
@axdanbol we'd like you to try your hand at resolving the base-href issue. I see we use base-href for both routes and assets. I believe we will need to separate how we think about them in order to make this work well for web components / apps that are served from CDN, but need nice route URLs at the deployment site. I'm thinking we let the base-href from html determine the route URL resolution (essentially no change in that area) whereas an input/web component attribute (`app-href`?) determines the base application href for assets that are delivered next to the application files on CDN.

axdanbol:
Implemented by #538

Issue #472, opened 2024-06-19
Landing page: Cosmetic & responsive design updates


LibbyUX:
The HRA Dashboards landing page has been adjusted to be more dynamic and cosmetic.

This is a low priority development request. 

Please see my [annotated image](https://www.figma.com/design/GAHPA5ni25Oq6qjeypvacx/Dashboards?node-id=2933-13249&t=wzQr96lzQZxOZmiu-4) of the updates and [new section of the MVP1 canvas](https://www.figma.com/design/GAHPA5ni25Oq6qjeypvacx/Dashboards?node-id=2923-23491&t=wzQr96lzQZxOZmiu-4) for all adjustments/breakpoints:

![image](https://github.com/hubmapconsortium/hra-ui/assets/97060856/6dd5827c-6e86-47ac-a22a-b8a54499519e)

bhushankhope:
@LibbyUX fixed in PR #477 . Thanks

bherr2:
@LibbyUX can you review and close if finished?

LibbyUX:
@bherr2 yes, thank you for the reminder ping!

Off the top, it does need to be closer to spec:
- Currently, the font on the visual cards does not match the spec. The live production cards use Nunito Sans, however, this was specified using our Label font, Metropolis 500/Medium.
- Diversity & Inclusion needs to be reordered still in navigation

Let me review this more thoroughly tomorrow morning. I think the footers may need to match the design system footer as well. Adding to my task sheet. Thank you so much!

bherr2:
@LibbyUX please review from this URL for now https://cdn.humanatlas.io/ui/dashboard-ui/

LibbyUX:
[Compared with notes](https://www.figma.com/design/GAHPA5ni25Oq6qjeypvacx/Dashboards?node-id=3005-7252&t=iU7XFYO84B1HLNgh-4). This is looking so fabulous!
- [Display font & overall notes](https://www.figma.com/design/GAHPA5ni25Oq6qjeypvacx/Dashboards?node-id=3005-8060&t=iU7XFYO84B1HLNgh-4)
- [Visual cards](https://www.figma.com/design/GAHPA5ni25Oq6qjeypvacx/Dashboards?node-id=3005-7308&t=iU7XFYO84B1HLNgh-4): Questions & Requests
[...]

Issue #470, opened 2024-06-18
Landing page: Cosmetic & responsive design updates


LibbyUX:
The CDE landing page has been adjusted to be more dynamic and cosmetic. This is a low priority development request. Please see my annotated image and [new section of the MVP1 canvas](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2147-18057&t=xYaOjtb3g7VmYGbF-4) for all adjustments/breakpoints:

![image](https://github.com/hubmapconsortium/hra-ui/assets/97060856/288e0f4e-6e5e-4364-b908-06cc7b6439bf)

LibbyUX:
Quick update for devs: The [app footer variant](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=2762-3953&t=waRA5FezxZfdwb4n-4), plus all of it's breakpoints, in the design system repo, has been updated to be more dynamic and aesthetic. Changes have been pushed to the CDE and Dashboards design files.

App footer updates for all breakpoints:
- Logo size adjusted
- Label font/sizes finalized (hopefully)
- Medical disclaimer text color updated
- Padding/spacing updated
- Icon buttons (social media + funders) updated

 I hope that this is the last app footer change for a while! I won't be updating it again unless I receive a modification request. Please let me know if there is anything more I can do for this one to make it easier for development. Thanks!

LibbyUX:
@edlu77, Thanks so much for the updates at https://667c32b1d316be6962072132--hra-ui.netlify.app/apps/cde-ui/ - We are super close! The responsive updates are amazing and, in general, it looks fabulous.

I compared Figma and the preview side-by-side in a new [design audit section](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2251-18162&t=9TwZbxqmg2N3cara-4).

I left my requests, questions, and screenshots in the [Notes section](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2256-18330&t=9TwZbxqmg2N3cara-4)

[...]

Issue #439, opened 2024-06-14
New HRA font styles


LibbyUX:
I'm inspecting the CDE landing on Google Chrome, and I see the following: 
![image](https://github.com/hubmapconsortium/hra-ui/assets/97060856/7c756194-0a29-42e0-9ac5-e97a88d00cf8)

This is what I see in Figma:
![image](https://github.com/hubmapconsortium/hra-ui/assets/97060856/492f5f5c-abf5-4647-aa0f-cd362f88cac5)

Does this match the spec? Also, how can I help make fonts/font weights easier for our team? I know it's been a moving target for us all as we begin using the new design system. Things are starting to get settled and you can see this progress in the Design System Repo: [Pattern Library Typography Style Sheet](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=3739-15606&t=FQZbf1z5LbpL8k6x-4)

LibbyUX:
While working on #452 , I supplied [design notes](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=4269-24264&t=hDK2XRTIVA3mJdUK-4) & side-by-side images while inspecting font styles online vs. in the Figma files. Please see my font design notes from 6.17.2024 for both the cde-ui, dashboards-ui, and all new/updated HRA interfaces.

![image](https://github.com/hubmapconsortium/hra-ui/assets/97060856/c2d91bac-d767-411d-9e27-e9db77db8d45)

edlu77:
Fonts updated in develop

PR #504, opened 2024-06-28
Fix font imports

edlu77:
Fixed font weight issues in CDE and HRA-dashboard 

#439
#470
#472


LibbyUX:
@bherr2 & @edlu77, 

Quick question as I get familiar with reviews: 

If I don't see a preview link, is there somewhere else on this page where I can review the changes in a preview format? Or should I be reviewing the code itself?

bherr2:
@edlu77 @LibbyUX I think nx may be having some server trouble right now

Issue #498, opened 2024-06-26
Use new /v1/session-token route for custom data sources


bherr2:
In the EUI when there are custom data sources set, use the /v1/session-token API to get a token and then give that token to db-status to start the EUI as normal. See <https://apps.humanatlas.io/api--staging/#post-/v1/session-token> for usage. This session-token route will be available in HRA-API v0.7.0 client.

bherr2:
You'll be able to check for data-sources and filters here: https://github.com/hubmapconsortium/hra-ui/blob/main/libs/ccf-shared/src/lib/services/data-source/api-endpoint.service.ts#L108-L109 grab the token from /v1/session-token and then set the "hubmapToken" via its output.

axdanbol:
Covered by #497

PR #517, opened 2024-07-02
Get a session token when custom data sources / filters used

edlu77:
#498


libs/ccf-shared/src/lib/services/data-source/api-endpoint.service.ts
```diff
@@ -101,11 +104,39 @@ function spatialSceneNodeReviver(nodes: RawSpatialSceneNode[]): SpatialSceneNode
   providedIn: 'root',
 })
 export class ApiEndpointDataSourceService implements DataSource {
+  private readonly http = inject(HttpClient);
```
bherr2, line 107: Hmm, this is ok to do for this PR, but when you replace the CCF-API client with the @hra-api/ng-client you'll want to use the equivalent function from there. I think the logic will be similar to below.

libs/ccf-shared/src/lib/services/data-source/api-endpoint.service.ts
```diff
@@ -101,11 +104,39 @@ function spatialSceneNodeReviver(nodes: RawSpatialSceneNode[]): SpatialSceneNode
...
+        tap(([sources, filter, endpoint]) => {
+          if ((sources && sources.length > 0) || filter) {
+            const sessionTokenRequest = {
+              dataSources: sources,
+              filter: filter,
+            };
+            console.log(sessionTokenRequest);
+
+            // Get session token
+            this.http.post(endpoint + '/session-token', sessionTokenRequest).subscribe((resp: DefaultParams) => {
```
bherr2, line 129: You'll want to catch errors here as the endpoint might not have /session-token and then you'll just need to set the filter to ''. May need to do that for other cases here as well.

libs/ccf-shared/src/lib/services/data-source/api-endpoint.service.ts
```diff
@@ -101,11 +104,39 @@ function spatialSceneNodeReviver(nodes: RawSpatialSceneNode[]): SpatialSceneNode
...
+  private readonly http = inject(HttpClient);
+
   constructor(
     private readonly api: DefaultService,
     private readonly globalConfig: GlobalConfigState<ApiEndpointDataSourceOptions>,
   ) {
     globalConfig.getOption('hubmapToken').subscribe(buster$);
+    combineLatest([
+      globalConfig.getOption('dataSources'),
+      globalConfig.getOption('filter'),
```
bherr2, line 116: Oh also, filter will change a lot as they use the tool. We need to capture the 'initial filter'. I'm not sure if we have logic in here to catch that.

apps/ccf-body-ui-wc/src/environments/environment.staging.ts
```diff
@@ -14,7 +14,7 @@ export const environment = {
     hubmapPortalUrl: 'https://portal.hubmapconsortium.org/',
     hubmapDataUrl: 'https://search.api.hubmapconsortium.org/v3/entities/search',
     hubmapAssetsUrl: 'https://assets.hubmapconsortium.org',
-    hubmapToken: localStorage.getItem('HUBMAP_TOKEN') ?? '',
+    token: localStorage.getItem('SESSION_TOKEN') ?? '',
```
bherr2, line 17: You can remove this kind of usage too. That was a debugging thing used long ago, but no longer relevant.

apps/ccf-eui/src/environments/environment.staging.ts
```diff
@@ -10,12 +10,8 @@ export const environment = {
   dbOptions: {
     ccfOwlUrl: 'https://apps.humanatlas.io/hra-api--staging/v1/ccf.owl.n3store.json',
     ccfContextUrl: 'https://hubmapconsortium.github.io/ccf-ontology/ccf-context.jsonld',
-    hubmapDataService: 'search-api',
-    hubmapDataUrl: 'https://search.api.hubmapconsortium.org/v3/entities/search',
-    hubmapToken: localStorage.getItem('HUBMAP_TOKEN') ?? '',
-
-    useRemoteApi: false,
-    remoteApiEndpoint: 'https://apps.humanatlas.io/api/v1',
+    token: localStorage.getItem('SESSION_TOKEN') ?? '',
```
bherr2, line 13: Won't need these

apps/ccf-eui/src/environments/environment.ts
```diff
@@ -12,14 +12,10 @@ export const environment = {
     'https://data.sennetconsortium.org',
   ],
   dbOptions: {
-    ccfOwlUrl: 'assets/ccf.owl.n3store.json',
+    ccfOwlUrl: 'https://apps.humanatlas.io/hra-api/v1/ccf.owl.n3store.json',
```
bherr2, line 15: When you remove usage of CCFDatabase this will be unnecessary as well

apps/ccf-eui/src/environments/environment.ts
```diff
@@ -12,14 +12,10 @@ export const environment = {
...
     ccfContextUrl: 'https://hubmapconsortium.github.io/ccf-ontology/ccf-context.jsonld',
```
bherr2, line 16: When you remove usage of CCFDatabase this will be unnecessary as well

libs/ccf-database/src/lib/ccf-database.ts
bherr2, line 1: This whole file should be removed, no need to try to update it

apps/ccf-eui/src/app/core/services/data-source/data-source.service.ts
```diff
@@ -21,9 +20,9 @@ export interface DelegateDataSourceOptions {
 })
 export class DelegateDataSourceService extends InjectorDelegateDataSourceService<DelegateDataSourceOptions> {
```
bherr2, line 21: We won't need a delegate data source service, we'll just use the ApiDataSourceService.

bherr2 (changes requested):
I know you are in the middle of this, but put in a few first comments. You'll be ripping out a lot more code soon.

bherr2:
image
 keep that negative number going up! A lot of code to remove 😄 

You should be able to remove the tripe-store-utils library as well.

apps/ccf-eui/src/app/app.component.ts
```diff
@@ -335,7 +335,7 @@ export class AppComponent implements OnInit {
    * Gets login token
    */
   get loggedIn(): boolean {
```
bherr2, line 337: This won't be possible/logical anymore. It'll just have to be set from outside (if at all)

apps/ccf-eui/src/app/core/services/data-source/worker-data-source.service.ts
bherr2, line 1: We won't need worker data source service as it used ccf database previously (and never quite worked)

libs/ccf-shared/src/lib/services/data-source/api-endpoint.service.ts
```diff
@@ -219,4 +203,28 @@ export class ApiEndpointDataSourceService implements DataSource {
       map((data) => (reviver ? reviver(data) : data)),
     );
   }
+
+  private checkForSources(): Observable<unknown> {
+    return combineLatest([this.globalConfig.getOption('dataSources'), this.globalConfig.getOption('filter')]).pipe(
+      debounceTime(1000),
```
axdanbol, line 209: 1000ms debounce seems a bit excessive. Let's work together to reduce/remove this

libs/ccf-shared/src/lib/services/data-source/api-endpoint.service.ts
```diff
@@ -219,4 +203,28 @@ export class ApiEndpointDataSourceService implements DataSource {
...
+        }
+      }),
+    );
+  }
+
+  private getSessionToken(params: SessionTokenRequestParams) {
+    this.api.sessionToken(params).subscribe((resp: DefaultParams) => {
+      const token = resp.token;
+      this.globalConfig.patchState({ token });
+      localStorage.setItem('SESSION_TOKEN', token ?? '');
```
axdanbol, line 226: Should we really save the token in localStorage? Do we ever read it from localStorage?

axdanbol (changes requested):
Some tweaking.
I'll play more with the previews tomorrow to check for bugs.

libs/ccf-shared/src/lib/services/data-source/api-endpoint.service.ts
bherr2: No we don't need to do that anymore

axdanbol (approved):

PR #518, opened 2024-07-02
show/hide empty fields and css updates on vis page

bhushankhope:
- add show/hide empty fields functionality
- update css of color picker rows
- update css of download png button

### DIFF
--- a/libs/cde-visualization/src/lib/models/metadata.ts
+++ b/libs/cde-visualization/src/lib/models/metadata.ts
@@ -21,4 +21,17 @@ export interface Metadata {
   creationDate?: string;
   /** Creation time */
   creationTime?: string;
+  /** Extra metadata for example datasets */
+  sampleExtra?: SampleMetadataExtra;
+}
+
+export interface SampleMetadataExtra {
+  /** Sample type, generally '2D' or '3D' */
+  type: string;
+  /** Organ name */
+  organ: string;
+  /** Data file url*/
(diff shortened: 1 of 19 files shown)


LibbyUX (changes requested):
Hi @bhushankhope - This is looking pretty good.

**Approved:**
- update css of color picker rows
- update css of download png button

Show/hide functionality looks good for sample visualizations.

I am requesting changes for user generated **add show/hide empty fields functionality**
- I created my own visualization and technology still showed
  - Note that I did add something to the input initially, however, I backspaced it out. The result was that it showed in the metadata.
- I am also hoping we can get this closer to specification for user-generated with the colors:

image

image

libs/cde-visualization/src/lib/models/metadata.ts
```diff
@@ -17,8 +17,19 @@ export interface Metadata {
...
+  sampleExtra?: SampleMetadataExtra;
+}
+
+export interface SampleMetadataExtra {
+  /** Sample type, generally '2D' or '3D' */
+  type: string;
+  /** Organ name */
+  organ: string;
+  /** Data file url*/
+  fileUrl: string;
```
axdanbol, line 32: Let's rename this to `sampleUrl` to prevent confusion

axdanbol (changes requested):
Don't forget to add code docs!

axdanbol (approved):

PR #529, opened 2024-07-08
hide download button in ftu-ui footer

bhushankhope:


axdanbol:
Some package install problem...
Minimatch seems to just have released a new version so maybe that has something to do with it.
I will restart the build tomorrow and it will hopefully work

Issue #521, opened 2024-07-03
Overlay-Scrollbar


LibbyUX:
# New component for all apps: Scrollbar appearing on hover
@axdanbol and I caught up yesterday on scrollbars. In the spirit of improving the density of our interfaces, I have an ask for development: Can we please implement a scrollbar that appears on hover of certain modules/containers? This is only for certain containers, not the overall scrollbar of the web page.

This component would work wonders instantly in FTU Explorer components (FTU Library, Matrix Component, and Source Data Section) and Cell Distance Explorer components (Cell Type list & Histogram Legend).

Eventually, we will need this component for the EUI, RUI, the new Digital Object Database in the HRA Portal, and ASCT+B Reporter. There's probably more UIs that will need this, too!

This component pattern is intuitive, clean, and it maximizes what we can show in our UIs.

You can see this functionality in several web browser apps, like Slack and Figma.

![FD47B1AC-3FE2-4FB2-8302-E211146F8E6F](https://github.com/hubmapconsortium/hra-ui/assets/97060856/bdac0da9-d738-44b7-a0cf-10abd234a1d6)

## Figma

I have specified this [scrollbar](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=4432-29127&t=3yTkV0MY1LkgR0Ln-4) in the Design System Repo. It has two variants, 12px & 8px. 

This design has been implemented in components in the redesigned CDE ([Cell Types list](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2228-16952&t=oDGeXCPTqaPIVvjk-4) & [Histogram legend](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2228-15687&t=oDGeXCPTqaPIVvjk-4)).

axdanbol:
Investigate https://kingsora.github.io/OverlayScrollbars/ for implementing this feature

axdanbol:
Or https://github.com/MurhafSousli/ngx-scrollbar

LibbyUX:
@axdanbol - Thanks so much for everything on this! Let me know if there's a better name for either of these atoms!

### File: Design System Component
- [Overlay-Fade](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=214-1697)
[...]

PR #543, opened 2024-07-10
Add scrollbar appearing on hover to cde cell types

edlu77:
#521

### DIFF
--- a/libs/cde-visualization/src/lib/components/cell-types/cell-types.component.scss
+++ b/libs/cde-visualization/src/lib/components/cell-types/cell-types.component.scss
@@ -79,9 +79,14 @@ $blue-theme: mat.define-light-theme(
   }
 
   .table-overflow-container {
-    overflow: auto;
     grid-area: table;
     height: 100%;
+    scrollbar-gutter: stable;
+    overflow: hidden;
+
+    &:hover {
+      overflow: auto;
+    }
 
     table {
       ::ng-deep {


LibbyUX (changes requested):
@edlu77 - Ok WHAT?!?! This is so exciting as I had no idea y'all were working on it. We're already so close!! Thank you so much for working on this component! This will be incredibly useful for all of our apps.

Today I'm on my Macbook, so this feedback might be specific to this OS. I reviewed the preview with the prototype side by side.

https://github.com/hubmapconsortium/hra-ui/assets/97060856/b63059ee-d50a-4b88-aafd-11e4fa9016f0

### Questions & notes
- Is it possible to remove the stroke/border track around the pill-like scrollbar?
- Can we reduce the opacity of the scrollbar? This might be helpful for super info-dense UIs later on. 
- The histogram legend scrollbar seems to always be available, but that might be because I was hovering over the entire Histogram component. Making a note for us on that!
- I started this review earlier this morning, got distracted, and could not replicate it for the review video just now: Initially I noticed a jumping in the width of the pill-like scrollbar while hovering over the Cell Types list. The only difference is that I was working on my laptop without being connected to my large monitor. Making a note that I saw that earlier!

### Resources
- [Design System Repo Scrollbar](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=4432-29127&m=dev)
Screenshot 2024-07-11 at 9 58 28 AM

axdanbol:
I'm closing this pr as the issue is much more complicated than @edlu77 might have first though. The solution needs to work for all scrollbars in an app and css alone doesn't provide the right tools. I've linked a library in the issue that seems promising but it needs further investigation.

Issue #499, opened 2024-06-26
Visualization app QoL updates


LibbyUX:
# Visualization app QoL updates
- Closing out major MVP1 work with cosmetic, responsive, and functionality updates for our new CDE!
- Please inspect all redesigned components carefully for responsive and aesthetic details! Any component appearing in the new canvas has been modified from our initial CDE.
- Redesigning this interface was extremely intentional to stay consistent and responsive, and to optimize screen space.
- Please don't hesitate to contact me with questions - I am always happy to help!!!
- Most of these updates are aesthetic/responsive, so I will highlight the UX updates needed below:

### UX updates + Figma links
- [x] Implement [new tooltip text](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2251-18072&t=9TwZbxqmg2N3cara-4) for the Cell Type module
- [x] Anchor icon: Implement a [tooltip hovercard](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2257-25191&t=9TwZbxqmg2N3cara-4) for this target point in the cell types list and legend

## Design repo resources
- [Typography Sheet](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=3739-15606&t=uNRJ0Klo3nHFxkfO-4)

## CDE design file
- Canvas: [Vis app redesign](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2203-11726&t=9TwZbxqmg2N3cara-4) 6.26.2024
- [Design changelog](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2251-18160&t=9TwZbxqmg2N3cara-4)
- [Updated components](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2251-17944&t=9TwZbxqmg2N3cara-4)
- [2D Sample](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=2203-13292&t=9TwZbxqmg2N3cara-4)
[...]

PR #548, opened 2024-07-12
Cde visualization app updates

edlu77:
#499 (excluding histogram styles)

### DIFF
--- a/apps/cde-ui/src/app/app.component.ts
+++ b/apps/cde-ui/src/app/app.component.ts
@@ -7,7 +7,7 @@ import { ScreenSizeNoticeComponent } from './components/screen-size-notice/scree
 /** Max width to show screen size notice */
 export const SCREEN_SIZE_NOTICE_MAX_WIDTH = 1280;
 /** Max height to show screen size notice */
-export const SCREEN_SIZE_NOTICE_MAX_HEIGHT = 832;
+export const SCREEN_SIZE_NOTICE_MAX_HEIGHT = 900;
 
 /**
  * App component for CDE
--- a/apps/cde-ui/src/app/components/screen-size-notice/screen-size-notice.component.html
+++ b/apps/cde-ui/src/app/components/screen-size-notice/screen-size-notice.component.html
@@ -6,7 +6,7 @@
 </h3>
 
 <mat-dialog-content class="content">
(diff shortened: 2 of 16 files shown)


LibbyUX (commented):
Visually we are there 100% on removing the corner radiuses, the strokes/borders, and the padding spacing! Amazing work @edlu77 !!

Iterative questions:
- Is it possible to make the containers more responsive?
  - On zoom, the header nav controls are pushed further to the right, but it would be amazing if they could squish in to the min width? Or are we already there?
  - Is it possible to make the vertical height behavior of the visualization module & histogram component behave more similarly to the spec?
  
This is new territory for me, so don't hesitate to let me know what is/isn't possible. I always zoom in/out on our interfaces to create UI annotation images with legible fonts at dimensions of 2816x1816. The EUI is a great example of how zooming in keeps everything contained in the browser window!

edlu77:
@LibbyUX All the styles on the page should be pretty close to the spec now. Still some additional tweaks needed for responsiveness but it's getting close!

LibbyUX (dismissed):
It's looking so good! Shoutout @edlu77 for REAL! Amazing job on this level of detail. These buttons look sooo good right aligned with the same padding in separate components! Thank you!!! 

image

I only caught one area for improvement and it is low priority. There is a slight jumpiness to the fonts between a certain breakpoint. We aren't optimized for this breakpoint anyway, but I found it interesting to see this slight jumpy text interaction while resizing my Google Chrome browser window!

D2BFD953-79B3-4F57-A03E-50B34738B4EE


I approve of this PR - thank you so much @edlu77 for working this. Our CDE is looking ultra fresh and user friendly!

edlu77:
@LibbyUX I see what happened here. The letter spacing isn't consistent between the different breakpoints...

I couldn't find the letter spacing value on Figma, what is it supposed to be?

LibbyUX:
@edlu77 nice catch! The letter spacing should be 0px, but we might need to iterate. Can you please let me know when the preview is updated, so that I can verify that it's good at 0px online?

Attaching our [UI Typography Style Sheet](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=3739-15606&t=WxpTpwlK0zgkVP8O-4) too! Talk soon!

axdanbol (approved):

Issue #583, opened 2024-07-25
Brand: Brandmark


LibbyUX:
# Brand: Brandmark
More information will be provided on our brandmark in the style guide. This is information about the brandmark component.
### Figma
- [Brand: Brandmark](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=82-776)

## Variant Usage
- Default color: Red
- 48px to 8px: Used for favicons, logo buttons in navigation components (headers, side nav, footers), and smaller logo variants
- Scalable: Not used often around the UIs, but available for sizes starting at 49px and above

PR #594, opened 2024-07-30
Design system brandmark storybook

edlu77:
#583


LibbyUX (commented):
This looks great!

Should I be reviewing if the component in Storybook is linked to the [Brandmark Figma canvas](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=82-776)? We will need to add it if so!

Mentioning @axdanbol for help answering this review workflow question, too!

Thanks so much!

image

edlu77:
@LibbyUX Thanks for catching that! I forgot to add in the link. WIll do that now.

libs/design-system/brandmark/src/lib/brandmark.component.html
```diff
@@ -0,0 +1 @@
+<img class="hra-brandmark" [attr.src]="svgPath() | assetUrl" alt="Human Reference Atlas brandmark" />
```
axdanbol, line 1: Use the `inlineSvg` directive from `ng-inline-svg-2` (already installed) to load the svg

libs/design-system/brandmark/src/lib/brandmark.component.scss
```diff
@@ -0,0 +1,3 @@
+:host {
+  display: block;
```
axdanbol, line 2: Add a css variable to control the color, `--hra-brandmark-color`. Default to `--sys-tertiary`

libs/design-system/brandmark/src/lib/brandmark.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  selector: 'hra-brandmark',
+  standalone: true,
+  imports: [CommonModule, AssetUrlPipe],
+  templateUrl: './brandmark.component.html',
+  styleUrl: './brandmark.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class BrandmarkComponent {
+  /** Brandmark color variant */
+  readonly color = input<string>('default');
```
axdanbol, line 18: Color will be controlled from css so please remove this input

libs/design-system/brandmark/src/lib/brandmark.component.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  /** Brandmark size to use */
+  readonly small = input<boolean>(false);
```
axdanbol, line 21: Since the brandmark is an svg it should automatically scale to whatever size we need so this input is not needed.

axdanbol (changes requested):
Please create a single svg that uses `currentColor` to control the color instead of multiple images

axdanbol (approved):

Issue #571, opened 2024-07-22
Icons: Material Symbols


LibbyUX:
# Icons: Material Symbols
We use [Material Symbols](https://fonts.google.com/icons?icon.style=Rounded&icon.set=Material+Symbols) around HRA interfaces.

![image](https://github.com/user-attachments/assets/f94a7140-0304-4ae4-b9aa-1ad9a005db01)

## Specifications
- Weight: 300
- Grade: 0
- Optical size: Almost always 24x24, however, include 20x20, just in case, for dense UIs and smaller components
- Style: Rounded

## Usage
- Used as standalone buttons to complete tasks
- Used with regular buttons to provide further visual context

![image](https://github.com/user-attachments/assets/5e9919bb-9d1e-43ae-90c8-db6e97d47e2f)

## Component
Our Material Symbol component is ever-evolving and iterative. It is updated almost every day with new additions (Figma variants) or improved organization. I pull these from the Material Symbols Figma plugin.
- Design System Repo: [Material Symbols](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=1790-54&t=7pLI9aOEH4FfwISb-4)

edlu77:
@LibbyUX Could you please make a page in the design system components for this issue/showcase

LibbyUX:
Yes, @edlu77 - Please see our new page, [Icons: Material Design](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=464-4)!

PR #598, opened 2024-07-31
Add material symbol component

edlu77:
#571


axdanbol:
I don't think this component makes a lot of sense to have. We can just use `<mat-icon>` directly. Instead we can create a standalone storybook file to show off `<mat-icon>`. @bherr2 What do you think?

bherr2:
> I don't think this component makes a lot of sense to have. We can just use `<mat-icon>` directly. Instead we can create a standalone storybook file to show off `<mat-icon>`. @bherr2 What do you think?

Sure, and of course you can still make a storybook to demonstrate the mat-icon usage (especially for all cases we support)

edlu77:
Updated to show all icons!

Issue #597, opened 2024-07-30
Filter Modal Window is Overcrowded


danqaurooni:
The filter modal in the EUI has become overcrowded and some checkboxes have become illegible. (See screenshot)

<img width="729" alt="Screenshot 2024-07-30 at 13 11 26" src="https://github.com/user-attachments/assets/e3d05783-62ed-4a5b-a0fe-1221127feac3">

OS: MacOS
Browser: Chrome

bherr2:
Yes, that's with the old EUI, but even the new EUI has graphical glitches that need resolved

![image](https://github.com/user-attachments/assets/29772829-edf9-4c52-b708-b8ec5c111e6d)

bherr2:
@LibbyUX anything more we can do to make this filter popup work? Ultimately need to rethink it completely, but tweaks to just make it work for now would be great.

LibbyUX:
@bherr2 is there a min/max width specified? I might recommend just having this area take up the whole screen if needed.

bherr2:
Looks like there are some bounds set. @edlu77 or @bhushankhope could tinker with it to see if we can get it to expand further when needed. Though that doesn't help with smaller screen sizes..

edlu77:
@bherr2 @LibbyUX I can work on this to make the popup more responsive to the contents. Some of the dimensions are currently hard coded which makes adding more stuff pretty annoying...

bherr2:
@edlu77 sounds good! Yeah, see what you can do to make it less annoying.

LibbyUX:
Thank you so much, @edlu77 & @bherr2! Please ping me if you need any guidance! 

Future versions goal: I will start brainstorming a component-based solution for this EUI filter feature. It needs to be easier for both our users and our team. Let me know if you have any components in mind already that could help us during the redesign of the EUI!

axdanbol:
@bherr2 Fix has been merged to develop

PR #602, opened 2024-08-01
EUI filter popup fixes

edlu77:
Tweaks to fix sizing issues in the EUI filter popup #597


LibbyUX (commented):
Thank you so much for these quick updates, @edlu77! Is it possible to make the filter flyout scroll at 200% zoom? Otherwise, I think we're good!

Screenshot 2024-08-01 at 10 16 14 AM (2)

edlu77:
@LibbyUX Scrolling should be enabled now. I also removed some filters in the index.html, so we can preview the updated component with all the items now.

axdanbol (approved):

Issue #574, opened 2024-07-22
Button: Mat-Sort-Header


LibbyUX:
# Button: Mat-Sort-Header
Please add an [Angular Material sort header](https://material.angular.io/components/sort/overview#using-sort-with-the-mat-table) to our HRA Design System Storybook. Note that I have made styling updates to align states/colors with Angular Material

![image](https://github.com/user-attachments/assets/3a610524-6fb3-41e1-a498-4d9dc60499b1)

## Styling Specifications
- Font: Label Medium - Metropolis 500px 14px/21px
- Material Symbols:
  - [Arrow Upwards Alt](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:arrow_upward_alt:FILL@0;wght@300;GRAD@0;opsz@24&icon.query=arrow+upward+alt&icon.size=24&icon.color=%235f6368&icon.style=Rounded)
  - [Arrow Downwards Alt](https://fonts.google.com/icons?selected=Material+Symbols+Rounded:arrow_downward_alt:FILL@0;wght@300;GRAD@0;opsz@24&icon.query=arrow+downward+alt&icon.size=24&icon.color=%235f6368&icon.style=Rounded)

## Behavior
- Angular's [mat-sort-header](https://material.angular.io/components/sort/overview#using-sort-with-the-mat-table) behavior is desired for this component. No custom behavior is needed. I will work to use fill widths/heights in table headers moving forward in Figma to make table headers behave more like Angular's.

## Usage
- Used to sort tabular data in mat-tables.
### Current UIs with this feature
- Cell Distance Explorer
- FTU Explorer

## Component
- Design System Component: [Button: Mat-Sort-Header](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=37-63)

PR #614, opened 2024-08-07
Design system storybook table

edlu77:
#574

Sort arrow is not actually using a mat-icon since it's actually an elaborate series of divs and is pretty hard to change. I tried to make it as close to spec as possible though.


LibbyUX (changes requested):
Thanks so much for getting this started, @edlu77 ! It's looking really good!

A request on this: Can we make the underline only appear on keyboard tab Focus state? Note that this is different from our focus visible state. Please see all state variants in the [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=2015-4422&m=dev). The state behavior that Angular provides in their [table with sorting example](https://material.angular.io/components/sort/overview#table-sorting) is exactly what I'm looking for!

Screenshot 2024-08-08 at 12 31 33 PM

edlu77:
@LibbyUX What's the difference in behavior between the focus visible state and keyboard tab focus state? Seems like they mean the same thing...

LibbyUX:
@edlu77, @axdanbol helped me understand the differences between focus and focus visible:
- https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#focus_vs_focus-visible
- https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#comparing_focus_and_focus-visible

I was naming components "focus" in Figma, when in reality, it was focus visible functionality I was trying to achieve. I am working to update all component naming conventions to focus visible when needed, along with specifying new (and actual) focus states for components that could use it.

edlu77:
@LibbyUX All right I see. Are we not using the red outline variant then?

LibbyUX:
@edlu77 for the sort header button, we will build both focus and focus-visible, please. 

Moving forward I will make sure the components I send to the Storybook design system use the proper naming convention for the CSS functionality they are trying to achieve. We will always have focus-visible for most buttons, but we may have focus states too, depending on how Angular is doing it.

I greatly appreciate your work on this one! We will soon be using this component a lot! 

Screenshot 2024-08-08 at 4 26 22 PM

libs/design-system/table/src/lib/table.component.ts
axdanbol, line 1: Rename this to `table-demo.component.ts` and move into a table-demo directory

libs/design-system/table/src/lib/table.component.html
```diff
@@ -0,0 +1,19 @@
+<table
+  mat-table
+  [dataSource]="dataSource || null"
```
axdanbol, line 3: Why `|| null`?

libs/design-system/table/src/lib/table.component.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Angular Material table with with sort feature
+ */
+@Component({
+  selector: 'hra-table',
+  templateUrl: 'table.component.html',
+  standalone: true,
+  imports: [MatTableModule, MatSortModule, CommonModule],
+})
+export class TableComponent implements AfterViewInit {
```
axdanbol, line 20: Rename to `TableDemoComponent`

libs/design-system/table/src/lib/table.component.ts
```diff
@@ -0,0 +1,74 @@
+import { CommonModule } from '@angular/common';
+import { AfterViewInit, Component, input, ViewChild } from '@angular/core';
+import { MatSort, MatSortModule, Sort } from '@angular/material/sort';
+import { MatTableDataSource, MatTableModule } from '@angular/material/table';
+
+/** Table data type */
+export interface TableData {
+  [name: string]: string | number;
```
axdanbol, line 8: No need for this to be generic in a demo component. Add `name` and `value` properties

libs/design-system/table/src/lib/table.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  dataSource = new MatTableDataSource(this.data());
+
+  /** Sorted data */
+  sortedData = this.data();
+
+  /** Columns in table */
+  columns: string[] = [];
+
+  /** Mat sort element */
+  @ViewChild(MatSort) sort!: MatSort;
```
axdanbol, line 34: Use `viewChild.required(...)` instead of decorator

libs/design-system/table/src/lib/table.component.ts
axdanbol: Also follow this example to simplify the component: https://material.angular.io/components/sort/overview#using-sort-with-the-mat-table

libs/design-system/table/src/lib/table.component.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  /** Sets data source and gets column names after view init */
+  ngAfterViewInit() {
+    this.dataSource.data = this.data();
+    this.dataSource.sort = this.sort;
```
axdanbol, line 39: Use an `effect` to set this in the constructor:
```ts
constructor() {
  effect(() => {
    this.dataSource.sort = this.sort();
  });
}
```

libs/design-system/table/src/lib/table.component.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+@Component({
+  selector: 'hra-table',
+  templateUrl: 'table.component.html',
+  standalone: true,
+  imports: [MatTableModule, MatSortModule, CommonModule],
+})
+export class TableComponent implements AfterViewInit {
+  /** Unsorted data */
+  data = input<TableData[]>([]);
```
axdanbol, line 22: Use `readonly` for properties that do not change

libs/design-system/table/src/lib/table-styles/table-styles.component.scss
axdanbol, line 1: Update px to rem

axdanbol (changes requested):

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  readonly data = input<TableDemoData[]>([]);
+
+  /** Table data source */
+  dataSource = new MatTableDataSource(this.data());
+
+  /** Sorted data */
+  sortedData = this.data();
+
+  /** Columns in table */
+  columns: string[] = [];
```
axdanbol, line 34: Just set the columns to `['name', 'value']`

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
```diff
@@ -0,0 +1,51 @@
...
+})
+export class TableDemoComponent implements AfterViewInit {
+  /** Unsorted data */
+  readonly data = input<TableDemoData[]>([]);
+
+  /** Table data source */
+  dataSource = new MatTableDataSource(this.data());
+
+  /** Sorted data */
+  sortedData = this.data();
```
axdanbol, line 31: Remove

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /** Columns in table */
+  columns: string[] = [];
+
+  /** Mat sort element */
+  sort = viewChild.required(MatSort);
```
axdanbol, line 37: readonly

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
axdanbol: Also readonly

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  templateUrl: 'table-demo.component.html',
+  standalone: true,
+  imports: [MatTableModule, MatSortModule, CommonModule],
+})
+export class TableDemoComponent implements AfterViewInit {
+  /** Unsorted data */
+  readonly data = input<TableDemoData[]>([]);
+
+  /** Table data source */
+  dataSource = new MatTableDataSource(this.data());
```
axdanbol, line 28: Readonly

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /** Sort data on load */
+  constructor() {
+    effect(() => {
+      this.dataSource.sort = this.sort();
+    });
+  }
+
+  /** Sets data source and gets column names after view init */
+  ngAfterViewInit() {
```
axdanbol, line 47: Remove

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
axdanbol: Initialize with the empty array `[]`

libs/design-system/table/src/lib/table-demo/table-demo.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  columns: string[] = [];
+
+  /** Mat sort element */
+  sort = viewChild.required(MatSort);
+
+  /** Sort data on load */
+  constructor() {
+    effect(() => {
+      this.dataSource.sort = this.sort();
+    });
```
axdanbol, line 43: Add effect to bind data:
```ts
effect(() => {
  this.dataSource.data = this.data();
});
```

libs/design-system/table/src/lib/table-demo/table-demo.component.html
```diff
@@ -0,0 +1,13 @@
+<table mat-table [dataSource]="dataSource" matSort aria-label="Table with sort function">
+  @for (column of columns; track column) {
```
axdanbol, line 2: Just create one column def for `name` and `value` respectively instead of a for loop

axdanbol (approved):

Issue #573, opened 2024-07-22
Button: Mat-Icon-Button


LibbyUX:
# Button: MatIconButton
To do: Please add our MatIconButton to Storybook:

- Figma: [Design System Components - Button: MatIconButton](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=24-876)
- Angular Material: [MatIconButton](https://material.angular.io/components/button/api#MatIconButton)

## Size variations
- Large
- Medium
- Small

## Component states
- Ready/default
- Hover
- Active/pressed
- Focus
- Focus Visible

## Behavior
Focus state overrides hover. Active/pressed overrides focus state.

## Usage
Used in dense application interface environments to achieve a specific task

## Icons
Material Symbols icons styled as specified, see:
- Styles: #571 

![image](https://github.com/user-attachments/assets/cbdc5725-1869-48eb-a89a-1490d29a4d8c)

Issue #557, opened 2024-07-16
Sidenav: HuBMAP Apps


LibbyUX:
# Implement HRA version of HuBMAP shared navigation component
- Design menu options to pitch tomorrow at CNS UX specs
- Redesign every app header to feature this button with shared nav functionality

# Resources:
- [Data Portal](https://portal.hubmapconsortium.org/): Use nav header button Resources and Apps icon to see menu
- HIVE UX-UI Collaboration Spreadsheet: [HuBMAP Tools Text & Image Assets](https://docs.google.com/spreadsheets/d/12obX2cWD6Sw3L0IWXMs47LT4YyBJEKbxNUCcNLqM98w/edit?pli=1&gid=0#gid=0)

<img width="917" alt="Screenshot 2024-07-16 at 2 16 18 PM" src="https://github.com/user-attachments/assets/e41e32c9-2fe1-4c4a-8776-a32563c68fdc">

LibbyUX:
# Status update:
- Moving all navigation areas to side nav for better mobile responsiveness. This feature will need to seamlessly fold into our existing and new side navs. Thank you @axdanbol for your help with this design!

### TO DO
Design updates for nav bars current UIs:
- HRA Portal current implementation (mobile side nav in use currently)
- Dashboards (mobile side nav in use currently)
- Cell Distance Explorer
- FTU Explorer
  
Passing this to Katy today for review on FTU Explorer & Atlas Apps.

LibbyUX:
### Design Review
Passed to @katyb tonight for feedback or approval:
- Prototype: [HuBMAP Shared Navigation (Side nav overlay)](https://www.figma.com/proto/VlNjSrctLEc0Wyu32FmnZq/Functional-Tissue-Unit-(FTU)-Explorer?page-id=5919%3A26286&node-id=5919-33586&viewport=742%2C-2884%2C0.29&t=YEO5HPNGEEkWyXGJ-1&scaling=contain&content-scaling=responsive&starting-point-node-id=5919%3A33586&show-proto-sidebar=1)

<img width="1280" alt="Screenshot 2024-07-18 at 10 31 06 PM" src="https://github.com/user-attachments/assets/a7b76d8f-f197-494c-ab33-e0158124404e">

<img width="918" alt="Screenshot 2024-07-18 at 10 32 21 PM" src="https://github.com/user-attachments/assets/6d2758b3-3acf-4a82-8281-8cec9bf1c13c">

[...]

PR #621, opened 2024-08-08
add nav header to storybook

bhushankhope:


LibbyUX (dismissed):
@bhushankhope - Thanks for getting this started! I know we just touched base in person, but following up on here for us and our colleagues:

- We will get the navbar color closer to spec
- We will add our apps icon button (#573) for the sidenav entry point (#557)

It's looking great so far, thanks for all of your help on this!

libs/design-system/nav-header/src/lib/nav-header.component.ts
```diff
@@ -0,0 +1,30 @@
+import { CommonModule } from '@angular/common';
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
+import { MatButtonModule } from '@angular/material/button';
+import { MatIconModule } from '@angular/material/icon';
+import { IconButtonSizeDirective } from '@hra-ui/design-system/icon-button';
+import { AppLogosComponent, AppLogosVariant } from '../../../logos/apps/src/index';
```
axdanbol, line 6: Need to update the location and tsconfig.base.json of app logos so it can be imported from `@hra-ui/design-system/app-logos` instead

libs/design-system/nav-header/src/lib/nav-header.component.ts
```diff
@@ -0,0 +1,30 @@
...
+  selector: 'hra-nav-header',
+  standalone: true,
+  imports: [CommonModule, MatButtonModule, MatIconModule, AppLogosComponent, IconButtonSizeDirective],
+  templateUrl: './nav-header.component.html',
+  styleUrl: './nav-header.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class NavHeaderComponent {
+  /** Variant of the header */
+  variant = input<AppLogosVariant>();
```
axdanbol, line 21: Add `'default'` as a default value to this input

libs/design-system/nav-header/src/lib/nav-header.component.html
```diff
@@ -0,0 +1,10 @@
+<hra-app-logos
+  [variant]="variant() ?? 'default'"
```
axdanbol, line 2: Add the default in the component's input instead

libs/design-system/nav-header/src/lib/nav-header.component.html
```diff
@@ -0,0 +1,10 @@
...
+  [appLink]="link()"
+  [appIcon]="icon()"
+  [appTitle]="title()"
+  [appDescription]="description()"
+></hra-app-logos>
+<button mat-icon-button hraIconButtonSize="large">
```
axdanbol, line 8: Should the icon size be affected by the variant?

libs/design-system/nav-header/src/lib/nav-header.component.scss
```diff
@@ -0,0 +1,8 @@
+:host {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 4.5rem;
+  padding: 0 0.75rem 0 1rem;
+  background-color: #eff2f5;
```
axdanbol, line 7: Colors should come from `sys-*` css variables

libs/design-system/nav-header/src/lib/nav-header.component.scss
```diff
@@ -0,0 +1,8 @@
+:host {
+  display: flex;
+  justify-content: space-between;
```
axdanbol, line 3: Prefer using a filler div element over space-between

axdanbol (changes requested):
I don't see the ng-package.json file or changes in tsconfig.base.json that marks this as a secondary entrypoint?

axdanbol (approved):

PR #623, opened 2024-08-09
update tree to collapse non active nodes

bhushankhope:


axdanbol (changes requested):
On mobile I am unable to scroll to the bottom of the menu when a large menu group is open: 
menu-scroll

bhushankhope:
> On mobile I am unable to scroll to the bottom of the menu when a large menu group is open: menu-scroll

Fixed this bug

axdanbol (approved):

PR #636, opened 2024-08-12
Add reference to updated SPARQL query

andreasbueckle:


bherr2:
Updated the PR with the new link and fixed numbers

Issue #646, opened 2024-08-14
Mat-Menu


LibbyUX:
# Mat-Menu
[Angular Material's Menu](https://material.angular.io/components/menu/overview) in HRA styles is activated by the "More" icon button. This menu contains mat-menu buttons #645 

The "more" icon button to mat-menu activation is a pattern to condense actions in information-dense UIs. New iterations of the FTU Explorer and CDE rely heavily on this component. 

### Figma
- Design System Components: [Mat-Menu](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=619-1552)
- Design System Repo: [Mat-Menu](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=5155-20512&t=cfcDHSYCf2SHDvRy-4)

![image](https://github.com/user-attachments/assets/24960581-7fdf-43ec-8a22-7fedcb34e493)

### FTU Explorer
![image](https://github.com/user-attachments/assets/03fb7d36-6f28-46c9-9dbb-24d0f5e52338)
### CDE
![image](https://github.com/user-attachments/assets/e7083805-e86d-4a16-97c2-6b7564272c6d)

LibbyUX:
Superseded by #781

Issue #645, opened 2024-08-14
Button: Mat-Menu


LibbyUX:
# Button: Mat-Menu
This button lives within [Angular Material Menus](https://material.angular.io/components/menu/overview) and has custom state styles. It contains Material Design icons and nested instances. The mat-menu button and mat-menus will be a common pattern in information dense interfaces. New iterations of the FTU Explorer and CDE rely on this component.

### Figma
- Design System Components: [Button: Mat-Menu](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=618-212)
- Design System Repo: [Button: Mat-Menu](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=3722-15717&t=cfcDHSYCf2SHDvRy-4)

![image](https://github.com/user-attachments/assets/0376766d-554c-4131-be06-36d880dc9890)

LibbyUX:
Superseded by #781

PR #650, opened 2024-08-16
Design system storybook menu

edlu77:
#646, #645


LibbyUX (commented):
This is perfect, thank you @edlu77 ! Do we have the medium and small variants as well? I can't find them currently, however, this large variant looks/feels perfect!

LibbyUX (approved):
Thank you so much @edlu77 ! Happy Monday!

axdanbol (changes requested):
Please refactor this story in the same way as table, etc.

axdanbol (approved):

Issue #660, opened 2024-08-20
Mat-Tree


LibbyUX:
# Mat-Tree
[Angular Material's Tree component](https://material.angular.io/components/tree/overview#padding-flat-tree-only) with HRA styles will be used in the new version of the FTU Explorer and the EUI. We will eventually need a nested tree for the EUI.

### Figma
- Design System Components: [Mat-Tree](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=786-4)
- Design System Repo: [Mat-Tree](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=4054-28184&t=9GRERRnoy6L2Req8-4)

![image](https://github.com/user-attachments/assets/475de5cc-0911-4684-a620-af9431202f69)

### FTU Explorer
- Figma: FTU Explorer iteration: App Page

![image](https://github.com/user-attachments/assets/b50488da-4d96-4706-b0c6-3455e94973a9)

LibbyUX:
Completed up to spec on Aug 21. Will submit an update request eventually for minor cosmetic updates.

PR #664, opened 2024-08-21
Add tree stories

edlu77:
#660


LibbyUX (commented):
Hi Edward, this is looking great. Thanks for getting this so close!

One question:
- Is it possible to situate the collapsed parent icon in the specified state?

Thanks so much!

image

edlu77:
@LibbyUX I've changed the icon, it should be using the correct one now.

Question about keyboard navigation, do we still want the same behavior as in the FTU Explorer?

LibbyUX:
Hi @edlu77, yes please to the same keyboard tab functionality with the focus visible state.

Thanks so much for knocking this out!

edlu77:
@LibbyUX Made some updates and the keyboard navigation should be working now. Let me know if there are any other changes needed!

LibbyUX (commented):
Good morning, @edlu77! Thanks for the work on this! The tab looks good.

I just noticed there is a blue outline after pressing, probably because of the specified focus state variant. 

Is it possible to distinguish focus vs focus visible on this component? If not, let's just remove the focus state and go with focus visible only. I will delete the focus variant from our Figma library, too. 

I designed this component thinking we would be able to have separate looks for focus and focus visible, after a conversation with @axdanbol. No worries if it isn't possible, just let me know! Thanks so much and happy Tuesday! 

72E1AD89-0DA3-44BF-9D88-1E345CE19688

libs/design-system/tree/src/lib/tree-styles/tree-styles.component.scss
```diff
@@ -0,0 +1,62 @@
+mat-tree {
+  --mat-tree-node-text-font: 'Metropolis';
```
axdanbol, line 2: The font values should come from variables. Set the value to `var(--sys-label-medium-font)` and `var(--sys-label-medium-weight)` for the weight

libs/design-system/tree/src/lib/tree-styles/tree-styles.component.scss
```diff
@@ -0,0 +1,62 @@
...
+  --mat-tree-node-text-weight: 500;
+  --mat-tree-container-background-color: white;
```
axdanbol, line 4: Use `var(--sys-primary-container)`

libs/design-system/tree/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './lib/tree-size/tree-size.directive';
+export * from './lib/tree.component';
```
axdanbol, line 2: Don't export the demo component

libs/design-system/tree/src/lib/tree.component.ts
axdanbol, line 1: Please rename this file and component to `TreeDemoComponent` (and the html file)

libs/design-system/tree/src/lib/tree-size/tree-size.directive.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+@Directive({
+  selector: '[hraTreeSize]',
+  standalone: true,
+  host: {
+    '[style.--mat-tree-node-text-size.rem]': 'fontSize()',
+    '[style.--mat-tree-node-min-height.rem]': 'nodeHeight()',
+    '[style.--mdc-icon-button-state-layer-size.rem]': 'nodeHeight()',
+    '[style.--mdc-icon-button-icon-size.rem]': 'iconSize()',
+    '[style.--node-margin.rem]': '0.5',
```
axdanbol, line 38: Please move constants to the tree-styles scss file

axdanbol (changes requested):
A few small tweaks

edlu77:
@LibbyUX Yes, that is the focus state. I agree it's a bit redundant since we already have a selected state which usually overrides it. You can only see this on expandable nodes that are clicked (since they can't be selected). I can remove the outline to make it cleaner.

axdanbol (approved):

PR #674, opened 2024-08-26
Design system form field input

bhushankhope:


LibbyUX (dismissed):
Excellent! Thank you @bhushankhope !

libs/design-system/input/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './lib/input-styles/input-styles.component';
```
axdanbol, line 1: No need to export this component

libs/design-system/input/src/lib/input-styles/input-styles.component.scss
```diff
@@ -0,0 +1,35 @@
+.mat-mdc-form-field {
+  font-family: 'Metropolis';
```
axdanbol, line 2: Read from a --sys variable. Add `-font` to get just the font family from the variables, i.e. `--sys-label-small-font`

libs/design-system/input/src/lib/input.component.stories.ts
```diff
@@ -0,0 +1,49 @@
...
+    }),
+  ],
+};
+export default meta;
+type Story = StoryObj;
+
+export const Primary: Story = {
+  args: {},
+  render: () => ({
+    template: `
```
axdanbol, line 25: Please split into 4 different stories

axdanbol (changes requested):

axdanbol (dismissed):

Issue #670, opened 2024-08-23
Breadcrumbs


LibbyUX:
# Breadcrumbs
Breadcrumbs help the user understand their location within websites. Breadcrumbs are used in Atlas Apps, CDE, and soon the HRA Portal.

### Building Blocks
- button/basic
- chevron right icon

## Resources
- [Design System Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=892-4)
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=1581-62&t=XX47KkPm5YWeGNCC-4)

<img width="997" alt="Screenshot 2024-08-23 at 9 32 39 AM" src="https://github.com/user-attachments/assets/6cc5516c-1de8-4f51-9a6e-f07b82eb89da">

## Examples
<img width="689" alt="Screenshot 2024-08-23 at 9 30 54 AM" src="https://github.com/user-attachments/assets/02f2fdf0-dff9-4de8-b26e-37f0b31e6963">

<img width="796" alt="Screenshot 2024-08-23 at 9 31 36 AM" src="https://github.com/user-attachments/assets/9be0484a-7080-4211-b06f-fb78fb462a02">

PR #682, opened 2024-08-29
Design system storybook breadcrumbs

edlu77:
#670


libs/design-system/breadcrumbs/src/lib/breadcrumbs/breadcrumbs.component.html
```diff
@@ -0,0 +1,10 @@
+@for (crumb of crumbs(); track crumb; let last = $last) {
+  @if (!last) {
+    <a [routerLink]="['', crumb.route]">
```
axdanbol, line 3: Use just `crumb.route` instead of `['', crumb.route]`

axdanbol (changes requested):
One tweak

Issue #679, opened 2024-08-28
Snackbar


LibbyUX:
# Snackbar
Snackbar in HRA style modifications inspired by:
- [Angular Material's Snackbar](https://material.angular.io/components/snack-bar/examples)
- [Material Design's Snackbar](https://m3.material.io/components/snackbar/overview)

## General notes
I designed many variants from the Material Design UI kit, but we are currently only using the single-line snackbar variant in the CDE. Soon we will use it in the HRA Portal, RUI, ASCT+B Reporter, and FTU Explorer.

## Single-line snackbar behavior
The default behavior is good for this component. Ideally, this snackbar moves the way it appears in Angular Material's examples. Additionally, please:
- Center at the bottom of the screen
- When snackbar is visible, it has 1 rem of space between the bottom of the browser window (or, whatever default they are using - it looks nice)
- Snackbars without close affordance are open for 6 seconds before they automatically disappear

### Resources
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Design-System-Repository?node-id=5662-21751&t=a3iAckVUrlkue9ft-4)
- [Design System Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Design-System-Components?node-id=1045-242)

![image](https://github.com/user-attachments/assets/96505b9d-2dc9-4f61-806e-b45b3e94e6a7)

PR #691, opened 2024-09-03
snackbar design system component

bhushankhope:


bhushankhope:
@LibbyUX I have created all stories for the snackbar, but I think we need to go over the design once, the gap in the snackbar with longer text is not consistent when the close button is visible. We can go over this once you are back. Thanks

LibbyUX (commented):
Hey @bhushankhope a couple of questions regarding #679 
- When does the snackbar auto-close? I might not be looking in the right place in the Storybook UI.
- Are we using the specified drop shadow for this component?

Thanks for working on this! It's looking good!

libs/design-system/snackbar/src/lib/snackbar.component.scss
```diff
@@ -0,0 +1,74 @@
...
+}
+
+::ng-deep .hra-snackbar-panel {
+  --mdc-snackbar-container-color: var(--sys-surface-container);
+  --mdc-snackbar-supporting-text-color: var(--sys-secondary);
+  max-width: 21.5rem;
+
+  .mat-mdc-snackbar-surface {
+    padding: 0;
+    box-shadow: 0px 5px 16px 0px #201e3d3d;
```
axdanbol, line 68: Update to use a --sys color, i.e. something like `rgb(from var(--sys-secondary) r g b / 0.04)`

libs/design-system/snackbar/src/lib/snackbar.component.stories.ts
```diff
@@ -0,0 +1,137 @@
...
+    <button mat-flat-button (click)="snackbar.open(message(), action(), close(), actionButtonPosition())">
+      Open Snackbar
+    </button>
+  `,
+})
+class SnackbarDemoComponent {
+  readonly message = input<string>('');
+  readonly action = input<string>('');
+  readonly close = input<boolean>(false);
+  readonly actionButtonPosition = input<ButtonPosition>('start');
```
axdanbol, line 24: Add a duration input

libs/design-system/snackbar/src/lib/snackbar.component.stories.ts
```diff
@@ -0,0 +1,137 @@
...
+import { SnackbarService } from './snackbar.service';
+import { provideDesignSystem } from '@hra-ui/design-system';
+import { ButtonPosition, SnackbarData } from './snackbar.component';
+
+@Component({
+  selector: 'hra-snackbar-demo',
+  standalone: true,
+  imports: [MatButtonModule],
+  template: `
+    <button mat-flat-button (click)="snackbar.open(message(), action(), close(), actionButtonPosition())">
```
axdanbol, line 15: Pass duration as the 5th argument in the config object. https://material.angular.io/components/snack-bar/api#MatSnackBarConfig

axdanbol: Maybe move the logic into a simple method to not bloat the template

libs/design-system/snackbar/src/lib/snackbar.component.stories.ts
```diff
@@ -0,0 +1,137 @@
...
+    message: {
+      type: 'string',
+    },
+    action: {
+      type: 'string',
+    },
+  },
+  args: {
+    message: 'test message',
+    action: 'test action',
```
axdanbol, line 41: Add a default duration of 5 seconds

axdanbol (changes requested):
Minor tweaks

LibbyUX:
@axdanbol @bhushankhope Perfect! Thank you both so much!

Regarding the drop shadow, what is the opacity?

This is what I have in Figma for our main drop shadow style: 
image

bhushankhope:
@LibbyUX I have fixed the issues. New preview will be ready. Please review [here](https://github.com/hubmapconsortium/hra-ui/pull/691#issuecomment-2327226185)

axdanbol (approved):

Issue #700, opened 2024-09-06
Step Indicator


LibbyUX:
Please add our aesthetic step indicator to Storybook. This is used in module components that have multiple steps, such as the CDE Create Vis application, or apps.humanatlas.io preview applications like US1 & US2.

### Figma
- [Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Explorer-Design-System-Repository?node-id=5857-22490&t=fnHSI1BOY7Y3txdo-4)
- [Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Explorer-Components?node-id=1213-4)

![image](https://github.com/user-attachments/assets/11ae38dc-e802-4c94-bf86-c12f832ce5dc)

![image](https://github.com/user-attachments/assets/964c5df7-0a52-46f7-aae2-a7588c7fa261)

PR #704, opened 2024-09-10
Add step indicator stories

edlu77:
#700


LibbyUX (dismissed):
Lovely, thank you @edlu77 !

axdanbol (changes requested):
Please create a new component for the step indicator. We don't want to have to copy the styles all over the place. It will also make it easier to tweak the layout

LibbyUX (approved):
Thanks so much, @edlu77 !

axdanbol (approved):

PR #707, opened 2024-09-11
create mat dialog in design system

bhushankhope:


LibbyUX (dismissed):
Thank you @bhushankhope !

libs/design-system/dialog/src/lib/notice/notice.component.scss
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  .mat-mdc-dialog-title::before {
+    display: none;
+  }
+
+  .mat-mdc-dialog-content {
+    --mdc-dialog-supporting-text-tracking: 0.5px;
+    padding: 0 2rem;
+    font: var(--sys-body-large);
```
axdanbol, line 35: Missing letter-spacing

libs/design-system/dialog/src/lib/notice/notice.component.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const dismiss = await screen.findByRole('button', { name: 'Dismiss' });
+    expect(dismiss).toBeInTheDocument();
+  });
+
+  it('Action button should be visible', async () => {
+    service.openNotice(TITLE, MESSAGE, {
+      label: 'Action Button',
+      callback: () => {
+        ('');
```
axdanbol, line 40: Woot! Better practice it to use jest.fn() when creating empty functions. By doing so you can also test that the passed function is called when the action button is clicked.
```ts
const callback = jest.fn();
service.openNotice(..., {
   ...,
   callback
});

...

await user.click(dismiss)

expect(callback).toHaveBeenCalledTimes(1);
```

libs/design-system/dialog/src/lib/notice/notice.component.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      },
+    });
+
+    const dismiss = await screen.findByRole('button', { name: 'Action Button' });
```
axdanbol, line 44: Rename variable to `actionButton` or something similar.

axdanbol (changes requested):
A few minor changes

libs/design-system/dialog/src/lib/notice/notice.component.scss
bhushankhope: --mdc-dialog-supporting-text-tracking: 0.5px;
is letter spacing

axdanbol (approved):

Issue #767, opened 2024-10-04
Use design system nav header


LibbyUX:
![image](https://github.com/user-attachments/assets/615e0a33-1dd9-4557-8241-4249ee0621f7)

Please use our updated design system nav header. See:
- #765 
- #764 
- #766 
- #763 
- #675 
- [Figma: CDE Header](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=3099-14887&t=G29kHwFe7SBGHCOl-4)

Issue #766, opened 2024-10-04
HuBMAP sidenav updates


LibbyUX:
# Updates
This sidenav been updated to be more consistent with the nav header style:
- Background fill color updated
- Sidenav apps buttons: State fill colors updated
- Dividers added to section categories
- Font sizes/colors updated
- Scrollbar locked to right side
- In Figma (not necessarily needed in Storybook):
  - Sidenav apps buttons: Text string variables added for Product Titles and Product Descriptions to reduce number of Figma components needed
  - Partner logos added
  - Added to Sidenav variant
    - Scaling note: All sidenavs will live in this component to design consistently moving forward

Please reach out if you have questions!

# Resources
- [Storybook: AppsSideNavDemoComponent](https://cdn.humanatlas.io/ui--staging/storybook/design-system/?path=/docs/appssidenavdemocomponent--docs)
- [HRA Components: Sidenav](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=75-669)
- [Design System Repo: Sidenav max width](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=4805-21726&t=OL1ruNgysgpR28y0-4)
- [Design System Repo: Sidenav min width](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=6137-35826&t=OL1ruNgysgpR28y0-4)
- [Design System Repo: Sidenav apps buttons](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=6287-32883&t=OL1ruNgysgpR28y0-4)
- #557 
- #586 

![image](https://github.com/user-attachments/assets/4b1ec01b-d5b2-41ba-b7a9-ac64e8da855f)

![image](https://github.com/user-attachments/assets/c39900fe-294c-4aaf-8e2b-c997c0c4c516)

Issue #765, opened 2024-10-04
Nav Header updates


LibbyUX:
![091646C4-BA82-4345-ADE6-F0CB9E73C9F2](https://github.com/user-attachments/assets/44929743-136f-4fae-ba8e-c1565cb50384)

# Ready for dev
The new design compacts a LOT more in less pixels for mobile devices. The goal of redesigning this component was to keep our designs consistent across UIs, completely responsive, yet easier to maintain.
- Nav Header cosmetics have been updated to be more dynamic:
  - There is no longer text/image resizing as the app buttons stay the same for all breakpoints. All padding is unified.
- We have one variant ready for development right now (the nested #764 ensures we only need one NavHeader for all apps)
  - Note: Eventually we will need 2 more variants for the HRA Portal. I will submit an issue for this update then. We will need a mobile and desktop breakpoint due to the categorical structure of the HRA Portal. My goal is to eventually only use one navigation header for all HRA webpages (HRA Portal, app landing pages, apps.humanatlas.io, etc.)
- Note: Please make sure #764 is knocked out first before working on this. This is a nested component which can be applied on an instance specific basic.

## Resources
- Storybook: [NavHeaderComponent](https://cdn.humanatlas.io/ui--staging/storybook/design-system/?path=/docs/navheadercomponent--docs)
- [Design System Repo: Nav Header](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=4696-21925&t=OL1ruNgysgpR28y0-4)
- [HRA Components: Nav Header](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=0-1)

### Associated Components
Design System Repo & associated open design system issues:
- #592 
- [Navcat Button](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=4180-37707&t=OL1ruNgysgpR28y0-4)
- #562 
- [Product Logos](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=2033-1383&t=OL1ruNgysgpR28y0-4)
- #675 
- [Nav Header Buttons](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=6132-22528&t=OL1ruNgysgpR28y0-4)
- #764  
[...]

Issue #764, opened 2024-10-04
App Button/Nav Header Button updates


LibbyUX:
![image](https://github.com/user-attachments/assets/3ff44b38-4710-41c9-b20b-5c179e8bde9c)

# Ready for dev
I updated our app buttons to be versatile and responsive for all products. The new design compacts a LOT in less pixels for nav headers on mobile devices.

## Resources
- Storybook: [AppLogosComponent](https://cdn.humanatlas.io/ui--staging/storybook/design-system/?path=/docs/applogoscomponent--docs)
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=6279-22915&t=OL1ruNgysgpR28y0-4)
- HRA Components: [Nav Header Buttons](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=1411-410&t=rltZpoXg3U5XprIP-4)
  - [Former design](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=1411-409&t=rltZpoXg3U5XprIP-4)

### Design System Repo: Associated Components
- [Nav Header](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=3325-7623&t=OL1ruNgysgpR28y0-4)
- [Nav Cat Button](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=4180-37707&t=OL1ruNgysgpR28y0-4)
- [Product Logos](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=2033-1383&t=OL1ruNgysgpR28y0-4)
- [Software status indicator](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=5444-20834&t=OL1ruNgysgpR28y0-4)

### Design changelog
- Submitted atom for dev: #675 
- Added/submitted atom for dev: Software status indicator #763 
- Added atom/behavior: Navcat button + open sidenav functionality
- Added behavior: Max width to contain product titles in the new nav headers

### Dev to do
- [x] Update the name of AppButtonComponent in Storybook to Nav Header Buttons, if possible
[...]

PR #778, opened 2024-10-11
Cde nav header updates

edlu77:
Nav header updates #767

Contains updates to design system:
- Nav Header updates #765
- Nav Header Button updates #764
- Product Logos #675
- Software Status Indicator #763
- Sidenav updates #766


axdanbol (changes requested):
Tweaks

axdanbol (approved):

PR #783, opened 2024-10-18
Create expansion panel

bhushankhope:


libs/design-system/button-toggle/src/lib/button-toggle-size/button-toggle-size.directive.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { Component } from '@angular/core';
+import { render, screen } from '@testing-library/angular';
+import { ToggleButtonSizeDirective } from './button-toggle-size.directive';
+
+describe('ButtonToggleSizeDirective', () => {
+  it('should apply the styles based on the directive', async () => {
+    @Component({
```
axdanbol, line 7: Move the component definition outside the it & describe blocks

libs/design-system/button-toggle/src/lib/button-toggle-size/button-toggle-size.directive.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+      template: `<div hraButtonToggleSize="large" data-testid="dir"></div>`,
+      imports: [ToggleButtonSizeDirective],
+      standalone: true,
+    })
+    class ButtonToggleComponent {}
+
+    await render(ButtonToggleComponent);
+
+    const directive = screen.getByTestId('dir');
+    expect(directive.getAttribute('hrabuttontogglesize')).toBe('large');
```
axdanbol, line 17: Do not test for this attribute. It is brittle

libs/design-system/button-toggle/src/lib/button-toggle-size/button-toggle-size.directive.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+    expect(directive.style.font).toBe('var(--sys-label-large)');
```
axdanbol, line 18: Use [getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) to get the proper value.

libs/design-system/expansion-panel/src/lib/expansion-panel-animations.ts
```diff
@@ -0,0 +1,16 @@
+import { animate, AnimationTriggerMetadata, state, style, transition, trigger } from '@angular/animations';
+
+/** Animation for the expansion panel */
+export const EXPANSION_PANEL_ANIMATION_TIMING = '225ms cubic-bezier(0.4,0.0,0.2,1)';
+
+/** Animation declaration for the expansion panel */
+export const matExpansionAnimations: {
```
axdanbol, line 7: Export the body expansion value instead of wrapping it in an object, i.e.
```ts
export const BODY_EXPANSION = trigger(...)
```

libs/design-system/expansion-panel/src/lib/expansion-panel.component.html
```diff
@@ -0,0 +1,46 @@
...
+    #accordionItem="cdkAccordionItem"
+    tabindex="0"
+    [attr.aria-expanded]="accordionItem.expanded"
+    [attr.aria-controls]="bodyId"
+    [expanded]="expanded"
+  >
+    <div class="header">
+      @if (!disabled()) {
+        <button mat-icon-button data-testid="toggle" (click)="accordionItem.toggle()">
+          <mat-icon class="material-symbols-rounded">
```
axdanbol, line 12: Do you need the rounded class? Isn't that the default?

axdanbol (changes requested):
A couple of tweaks

libs/design-system/button-toggle/src/lib/button-toggle-size/button-toggle-size.directive.spec.ts
axdanbol: Actually ignore this one!

axdanbol (approved):

Issue #712, opened 2024-09-16
Workflow Card


LibbyUX:
# Workflow Card
Please add our Workflow Card to the design system. This component appears when users are completing a workflow process, such as creating a visualization in the CDE, or predicting cell type populations or tissue origins in apps.humanatlas.io preview applications.

### Determinate progress bar needed
Note: Cards with file uploads need determinate progress bars, so users can see the status of their file upload. See #622 for this component.

## Figma
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Explorer-Design-System-Repository?node-id=5517-21800&t=PGgdVBJSN0RcHOcM-4)
- [Explorer Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Explorer-Components?node-id=1286-4)

### Design System Repo with placeholder
![image](https://github.com/user-attachments/assets/d7451fc3-6fe5-4246-9580-85d26795cc7d)

### apps.humanatlas.io/us1
![image](https://github.com/user-attachments/assets/1a00ef75-99ee-4a93-ac7e-1d99bb9d0d05)

### CDE Create Visualization
![image](https://github.com/user-attachments/assets/5a935f9d-f0fa-41f7-bf8f-c0bff85743be)

PR #773, opened 2024-10-08
Cde design system

edlu77:
Added design system and updated styles for landing page, create vis page, and visualization page

Issue #731, opened 2024-09-20
Delete File Button


LibbyUX:
# Delete File Button
Please add Delete File Button inspired by an Etsy delete file UI, to the HRA Design System Storybook. It is needed for 3 recently approved UIs, along with more UIs in the future, such as the RUI and ASCT+B Reporter.

## Behavior
- This button appears after a user has uploaded a file and reports the file name name/file type to the user.
- This button gives users an option to delete their file and if they upload the wrong file.
- Currently, it only is used within Workflow Cards 
  - #712 
- It uses our icon button and looks like this on hover
![image](https://github.com/user-attachments/assets/3af5c340-53f6-445a-88db-364ee2e0af8c)

### Figma
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Explorer-Design-System-Repository?node-id=5676-22770&t=X0D1vCOZyGiIyp9H-4)
- [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=1333-193)

![image](https://github.com/user-attachments/assets/9ee697e3-7688-42d6-81a1-57fd33c17e8e)

PR #788, opened 2024-10-23
Design system workflow card

### LINKED ISSUE #622: Determinate progress bar
Katy noted that the CDE took a while to load in the sample data. Progress indicators for overall apps are needed. Bruce mentioned that development may be able to return a percentage. I also need to look into the subtle flashing block loading UI: This is a super common modern app loading pattern.

image

edlu77:
#712

- Uses determinate progress bar #622


edlu77:
@LibbyUX Please review and let me know if everything looks good. I've added a progress bar to the data and color map upload cards in the create vis page, so you can test that out as well.

LibbyUX (commented):
Thank you so much @edlu77!! This is looking and feeling so good!

Just a few, hopefully, small requests: 
- For #731 - Is it possible to add a hover state for the icon button on the left side? This whole interaction feels amazing, great work!
- I can't read the tooltips at the moment - The workflow card text appears above it
- Our inputs/selects have been a moving target this year but we are so close to finalized. I see that in the preview, the number input field has space between the label and the input text. Is it possible to reduce this space to 0 and align it as specified?
  - I am working to use [Material Design's text field component framework](https://m3.material.io/components/text-fields/specs) for all selects and inputs in a new component being called "[Text field v2](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=6542-24444&t=LJjdVuvcZEg9txVF-4)" in the Design System Repo. It's nearly identical to what was originally specified, but it is more usable in Figma. It also removes the size options, as that was not necessary and goes against what Angular and Material are doing. I will submit this update soon for our design system.

image

image


This looks and feels fantastic, thank you for all of your work on this flow, @edlu77!

LibbyUX:
Also @edlu77 I just want to mention that the progress bar is exquisite!!! Thank you!

edlu77:
@LibbyUX A lot of these issues are addressed in PR #773, which has the updated Delete File Button component and updates to color and styling in the app. It would probably make sense to import some of those changes here though.

LibbyUX:
@edlu77, thanks for pointing me in the right direction for #773. The delete button hover state and the tooltip cards look great over there.

Also, please disregard anything about the text field inputs/selects at the moment! I will submit a design system update when this is ready, so it's irrelevant for this review right now.

Per our conversation just now at my desk, I'll wait for you to ping me when this one is ready for review again!

Thanks so much for everything!

axdanbol (approved):

Issue #730, opened 2024-09-19
Info Modal


LibbyUX:
![image](https://github.com/user-attachments/assets/c5901a4a-260e-4a61-91a9-cda62267dfe2)

# Info Modal: Approved & Ready for Dev
Please add our new info modal to the Design System Storybook. This component reports all available information about specific areas of a dataset.

### Behavior
- These modals open on click of an interactive dataset button:
  - FTU-UI & FTU-UI-Small-WC: On click on a dataset in the matrix
  - CDE: On click of a node in the visualization inspect mode

We have two different styles currently approved:
- CDE, center aligned
- FTUE, tabular

## Figma
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Explorer-Design-System-Repository?node-id=5756-23245&t=TKx0gfqZQ8ctOHR4-4)
- [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Explorer-Components?node-id=1329-3362)
- [FTU Explorer Y3](https://www.figma.com/design/VlNjSrctLEc0Wyu32FmnZq/Functional-Tissue-Unit-(FTU)-Explorer?node-id=6318-26506&t=y0pdZMvsQYeZrJr3-4)
- [CDE Y3](https://www.figma.com/design/G0xVFFsAAnBwzHd8PtQEeb/Cell-Distance-Explorer?node-id=3019-20732&t=7sDYSFrbe91wdF8z-4)

PR #794, opened 2024-10-29
Create info modal component with two variants

edlu77:
#730


LibbyUX (commented):
Amazing, @edlu77 ! One quick question: Is it possible to use the icon button for the close modal button? It would be nice to see states on it.

Please let me know and I can review again after I know more. Otherwise, this looks perfect!

image

edlu77:
@LibbyUX Close icon has been replaced with icon button!

LibbyUX (approved):
Thanks so much @edlu77 !

axdanbol (approved):

Issue #813, opened 2024-11-15
Add new social logo icons


LibbyUX:
Please add a new icon set to our design system: Icons - Social Logos. All logo icons are sourced from the MingCute open-source library:
- https://www.mingcute.com/
- https://github.com/Richard9394/MingCute?tab=Apache-2.0-1-ov-file#readme
- https://www.apache.org/licenses/LICENSE-2.0

### Figma
- [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=1711-8329&t=jfZe78D5K1IP83V4-4)
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=6955-26116&t=0VHZ3NTwsATd5hrw-4)

![image](https://github.com/user-attachments/assets/01911e2f-ff81-4d54-8e58-81508df34d7d)

PR #895, opened 2025-01-13
Add new social media icons to design system

edlu77:
#813


libs/design-system/social-media-button/src/lib/social-media-button.component.ts
```diff
@@ -3,7 +3,7 @@ import { ChangeDetectionStrategy, Component, computed, input } from '@angular/co
 import { MatIconModule } from '@angular/material/icon';
 
 /** Social media name type */
-export type SocialMediaName = 'x' | 'facebook' | 'instagram' | 'youtube' | 'linkedin' | 'email';
+export type SocialMediaName = 'x' | 'facebook' | 'instagram' | 'youtube' | 'linkedin' | 'github';
```
axdanbol, line 6: We still want to have email as a social media button. You'll have to use an if/else block in the html to do it, i.e.
```
@if (name() === 'email') {
  <mat-icon>email</mat-icon>
} @else {
  <mat-icon [svgIcon]="'social:' + name()"?</mat-icon>
}
```

axdanbol (changes requested):

axdanbol (approved):

Issue #718, opened 2024-09-16
UI Images


LibbyUX:
# UI Images
We need a way to update UI images effortlessly across our interfaces, especially during the cosmetic redesign phase. We need one source of truth for all UI images in the design system. I will submit issues moving forward when these images are updated.

apps.humanatlas.io has two components, UI Section (#719) & component card (#722), that uses these images. I can see this image component scaling over into humanatlas.io eventually, too.

This component is needed for:
- #719 
- #831 
- #715 

## Figma
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/Explorer-Design-System-Repository?node-id=6060-22059&t=PGgdVBJSN0RcHOcM-4)
- [Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/Explorer-Components?node-id=1297-5526)

![image](https://github.com/user-attachments/assets/85fe06c8-8287-4383-ac3d-10032a23ddac)

axdanbol:
Images has been added in the design system storybook

PR #896, opened 2025-01-14
Added UI images story #718

gauri2029:
Added png images and a new story for UI images.


axdanbol (approved):
I've checked the preview and it works fine! I'll approve and merge since there isn't really anything else that is required by this issue.

PR #899, opened 2025-01-15
Deps upgrade

axdanbol:


axdanbol:
Need to wait until jest/jsdom is updated to support newer css features used by angular to get testing to work

PR #902, opened 2025-01-17
Adding web component card to design system

gauri2029:


libs/design-system/styles/_typography.scss
```diff
@@ -69,6 +69,8 @@ $overrides: (
 
   body-large-tracking: 0.0313rem,
 
+  body-large: 400 1rem / 1.5rem $_plain,
```
axdanbol, line 72: This is already the default for body-large so there is no need to override it. Please undo this change

libs/design-system/web-component-card/src/lib/web-component-card.component.scss
```diff
@@ -0,0 +1,59 @@
+:host {
+  display: block;
+  position: relative;
+  border-radius: 0.5rem;
+  overflow: hidden;
+  background-color: var(--sys-surface-container-low);
+  box-shadow: 0px 5px 16px 0px rgba(from var(--sys-shadow) r g b / 0.24);
```
axdanbol, line 7: Use `rgb`. `rgba` is a legacy alias according to mdn (https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#syntax)

libs/design-system/web-component-card/src/lib/web-component-card.component.html
```diff
@@ -0,0 +1,19 @@
...
+  <h1 *ngIf="webComponentName()" class="webComponentName">{{ webComponentName() }}</h1>
+  <span class="description">{{ description() }}</span>
+</div>
+
+<div class="cardFooter">
+  <button mat-button hraPrimaryButton hraButtonSize="medium">Use App</button>
+  <button mat-flat-button hraPrimaryButton hraButtonSize="medium">Embed</button>
+</div>
+
+<div class="content">
```
axdanbol, line 17: This component doesn't have any projected content so please remove this.

libs/design-system/web-component-card/src/lib/web-component-card.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="image">
+  <img [src]="imageUrl()" />
+</div>
+<div class="cardContent">
+  <h2 class="productTitle" [ngClass]="{ default_style: !webComponentName(), style_change: webComponentName() }">
```
axdanbol, line 5: Use an `@if` block to toggle between the two behaviors, i.e.
```html
@if (webComponentName()) {
  <span>{{ productTitle() }}</span>
  <h1>{{ webComponentName() }}</h1>
} @else {
  <h1>{{productTitle()}}</h1>
}
```

Then you won't need to use ngClass either

libs/design-system/web-component-card/src/lib/web-component-card.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="image">
```
axdanbol, line 1: There is no need to wrap the image element inside a div. You are setting the height of the image already so the extra div doesn't do anything anyways

libs/design-system/web-component-card/src/lib/web-component-card.component.scss
axdanbol, line 1: For future reference we use kebab-case when naming html classes, i.e. `card-footer` rather than `cardFooter`. Please update to follow our naming convension.

libs/design-system/web-component-card/src/lib/web-component-card.component.scss
```diff
@@ -0,0 +1,59 @@
...
+      width: 100%;
+    }
+  }
+
+  .cardContent {
+    padding: 1.5rem 1.5rem 0 1.5rem;
+  }
+
+  .default_style {
+    font: var(--sys-title-large);
```
axdanbol, line 22: Please set letter-spacing along with the font. `letter-spacing: var(--sys-title-large-tracking)`

libs/design-system/web-component-card/src/lib/web-component-card.component.scss
```diff
@@ -0,0 +1,59 @@
...
+    line-height: 2.25rem;
+    margin: 0;
+    margin-top: 0.5rem;
+    color: var(--sys-on-background);
+  }
+
+  .description {
+    margin-top: 0.5rem;
+    overflow-wrap: break-word;
```
axdanbol, line 45: Please remove. We should not be breaking text in the middle of words

libs/design-system/web-component-card/src/lib/web-component-card.component.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { WebComponentCardComponent } from './web-component-card.component';
+import { render, screen } from '@testing-library/angular';
+
+describe('WebComponentCardComponent', () => {
+  it('Error should be visible in the indicator', async () => {
```
axdanbol: This text could be improved. It should read as a complete sentence, for example: `it('should display the provided title, component name and description', ...)`
(Feel free to use the example text as is)

libs/design-system/web-component-card/src/lib/web-component-card.component.ts
```diff
@@ -0,0 +1,22 @@
+import { CommonModule } from '@angular/common';
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
+import { ButtonModule } from '@hra-ui/design-system/button';
+
```
axdanbol, line 4: Please add a doc comment to the component. Doc coverage will likely fail otherwise

axdanbol (changes requested):
Mostly tweaks and a little bit of cleanup needed

libs/design-system/web-component-card/src/lib/web-component-card.component.html
```diff
@@ -0,0 +1,19 @@
...
+  <img alt="image" [src]="imageUrl()" />
```
axdanbol, line 2: Please set `alt=""`. The image is purely presentational and the text `"image"` doesn't convey any information to the user about what it actually is showing. https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt#decorative_images

axdanbol (changes requested):
Please update alt attribute

libs/design-system/web-component-card/src/lib/web-component-card.component.html
```diff
@@ -0,0 +1,15 @@
+<img alt="" [src]="imageUrl()" />
+<div class="card-content">
+  @if (webComponentName()) {
+    <h2 class="style_change">{{ productTitle() }}</h2>
```
axdanbol, line 4: Please change class to `product-title`

libs/design-system/web-component-card/src/lib/web-component-card.component.html
```diff
@@ -0,0 +1,15 @@
...
+    <h1 class="web-component-name">{{ webComponentName() }}</h1>
+  } @else {
+    <h2 class="default_style">{{ productTitle() }}</h2>
```
axdanbol, line 7: Please change class to `product-title`

libs/design-system/web-component-card/src/lib/web-component-card.component.scss
```diff
@@ -0,0 +1,58 @@
...
+    margin-top: 0.5rem;
+    color: var(--sys-on-background);
+  }
+
+  .description {
+    margin-top: 0.5rem;
+    display: block;
+    width: 100%;
+    font: var(--sys-body-large);
+    letter-spacing: 0.0313rem;
```
axdanbol, line 48: Change value to `var(--sys-body-large-tracking)`

libs/design-system/web-component-card/src/lib/web-component-card.component.scss
```diff
@@ -0,0 +1,58 @@
...
+  img {
+    height: 15rem;
+    width: 100%;
+  }
+
+  .card-content {
+    padding: 1.5rem 1.5rem 0 1.5rem;
+  }
+
+  .default_style {
```
axdanbol, line 18: Replace `.default_style`, `.style_change`, and `.web-component-name` with the following:
```scss
.product-title,
.web-component-name {
  font: var(--sys-title-large);
  letter-spacing: var(--sys-title-large-tracking);
  color: var(--sys-on-background);
  margin: 0;
}

.product-title:has(+ .web-component-name) {
  font: var(--sys-title-medium);
  letter-spacing: var(--sys-title-medium-tracking);
  color: var(--sys-inverse-surface)
  margin-bottom: 0.5rem;
}
```

This uses the [`:has()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) selector to detect when both a product title and component name is present.
Also `line-height` is automatically set by `font: ...` so you don't need to set it yourself

axdanbol (changes requested):
A little bit of scss improvements

axdanbol (approved):

PR #922, opened 2025-01-24
EUI cosmetic update

axdanbol:


edlu77:
@LibbyUX We merged a bunch of changes, can you rereview to make sure everything still looks good?

edlu77:
> Hi @edlu77 - It's looking so good! Just one minor cosmetic update, if possible:
> 
> * Could we tweak the biomarker button toggle buttons to fill the space of the button toggle bar? I did a spatial search and noticed the biomarkers were like this after I had applied my spatial search filter.
> 
> image

@LibbyUX Thanks for catching that! It's fixed now.

bherr2:
> Does each card always have only one route? Or do certain cards have links to more than one route?

Yeah, one card (donor, tissue block, or tissue section) always has only one link

PR #935, opened 2025-01-30
Update team.yaml

LibbyUX:
Verbiage update: Change heading label text "References" to "Publications", per Katy, during CNS UX Specs meeting on 1.29.2025


axdanbol:
Will be addressed in #940

PR #944, opened 2025-01-31
EUI organ select update

edlu77:


apps/ccf-eui/src/app/app.component.ts
```diff
@@ -335,4 +291,37 @@ export class AppComponent implements OnInit {
...
+  readonly searchTextField = computed(() => this.organs().length.toString() + ' Organs Visible');
+
+  remove(organ: string): void {
+    this.organs.update((organs) => {
+      const index = organs.indexOf(organ);
+      if (index < 0) {
+        return organs;
+      }
+
+      organs.splice(index, 1);
```
axdanbol, line 312: This will modify the original array. Judging be the next line I believe your intention was something like:
```ts
const copy = [...organs];
copy.splice(index, 1);
return copy;
```

apps/ccf-eui/src/app/app.component.html
```diff
@@ -220,7 +220,37 @@
       <span>Test</span>
     </mat-menu>
   </hra-expansion-panel-actions>
-  <hra-expansion-panel-header-content> </hra-expansion-panel-header-content>
+  <hra-expansion-panel-header-content>
+    <form>
```
axdanbol, line 224: Please use a form control instead of `ngModel` etc. We can work together to get it working correctly. Also we might want to move this organ selection into it's own component.

axdanbol (changes requested):
Let's work together to address the requested changes!

axdanbol (approved):

PR #957, opened 2025-02-07
Develop

### LINKED ISSUE #954: All routes in CDE HuBMAP apps menu show 404 not found
In the CDE's HuBMAP apps menu, none of the routes are functioning. I checked in both Firefox and Chrome.

In every instance, `https://cdn.humanatlas.io/ui/cde-ui` shows before the correct route for each button.

Image

axdanbol:
Closes #954

### DIFF
--- a/libs/design-system/nav-header/src/lib/nav-header.component.html
+++ b/libs/design-system/nav-header/src/lib/nav-header.component.html
@@ -16,7 +16,7 @@
               [icon]="card.icon"
               [tagline]="card.name"
               [description]="card.description"
-              [link]="card.link | assetUrl"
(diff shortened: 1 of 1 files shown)


libs/design-system/nav-header/src/lib/nav-header.component.html
```diff
@@ -16,7 +16,7 @@
               [icon]="card.icon"
               [tagline]="card.name"
               [description]="card.description"
-              [link]="card.link | assetUrl"
```
bherr2, line 19: That's exactly what I thought the issue would be 😄. Thanks for getting it fixed!

Issue #953, opened 2025-02-05
Question: How can we programtically set "Consortium" in RUI metadata


maxsibilla:
We have noticed that in the metadata panel there is now a "Consortium" field that can be populated with either SenNet or HuBMAP. We want to know how we can programmatically populate this field from the SenNet Data Sharing Portal and the HuBMAP Ingest UI.

![Image](https://github.com/user-attachments/assets/bf4b8f44-0ff1-48f4-b024-fc385b1ce21c)

LibbyUX:
Hi @maxsibilla, thank you so much for this question! I am mentioning @bherr2 and @axdanbol for this discussion.

All, please let me know if any design work is needed here. Happy Wednesday!

bherr2:
@axdanbol can you check if the consortium data can be set programmatically like we do with first/last name, etc?

axdanbol:
@bherr2 I've checked and there is currently no way to set it programmatically. Would you like me to implement this feature? (It should be trivial)

bherr2:
Yes please!

PR #968, opened 2025-02-20
Rui consortium programatic input

axdanbol:
Also updated the create modal to handle programatically set inputs


axdanbol (commented):
Closes #953

PR #980, opened 2025-02-28
Added URLs for Jan Jensen

Hrishikesh-3459:


.vscode/settings.json
```diff
@@ -31,5 +31,6 @@
     "lib:utils",
     "lib:webcomponents",
     "lib:state"
-  ]
+  ],
+  "liveServer.settings.port": 5501
```
bherr2, line 35: @Hrishikesh-3459 please revert the change to .vscode/settiongs.json

PR #990, opened 2025-03-05
Add progress bar on page load

edlu77:


apps/ccf-eui/src/app/core/store/data/data.state.ts
axdanbol, line 1: Why is `statusMessage` being removed?

apps/ccf-eui/src/app/app.component.ts
```diff
@@ -113,10 +113,8 @@ export class AppComponent implements OnInit {
     return navigator.userAgent.indexOf('Firefox') !== -1;
   }
 
-  /** Emits true whenever the overlay spinner should activate. */
-  readonly spinnerActive$ = this.data.state$.pipe(map((state) => state?.status !== 'Ready'));
-
-  readonly loadingMessage$ = this.data.state$.pipe(map((x) => x?.statusMessage));
+  /** Emits true whenever the progress bar should activate. */
+  readonly progressBarActive$ = this.data.state$.pipe(map((state) => state?.status !== 'Ready'));
```
axdanbol, line 117: I don't think this will be enough to determine when the progress bar should show. This only checks the database status but you'll have to show the progress bar when there are pending requests such as filter, etc. Lets talk to Bruce to determine which events contribute to this

axdanbol (changes requested):

PR #1004, opened 2025-03-21
Add card-info-component

edlu77:


apps/humanatlas.io/src/index.html
```diff
@@ -7,23 +7,6 @@
     <link rel="icon" type="image/png" href="favicon.png" />
     <link rel="preconnect" href="https://fonts.gstatic.com" />
```
axdanbol, line 8: Remove the preconnects

apps/humanatlas.io/src/styles.scss
```diff
@@ -1,6 +1,7 @@
 @use 'light-theme';
 @use 'theming';
 @use 'utils';
+@import url('https://fonts.googleapis.com/css2?family=Material+Symbols+Rounded:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200');
```
axdanbol, line 4: This is not correct. Rounded symbol is already included in the theming. If you need a different optical size apply
```scss
.the-icon-class {
  font-variation-settings:
  'FILL' 0,
  'wght' 300,
  'GRAD' 0,
  'opsz' 24 /* <- Optical size goes here! */
}
```
inside the count-info component instead!

axdanbol (changes requested):

axdanbol:
We need to update the icon sets to use variable fonts: https://fontsource.org/docs/getting-started/material-symbols

axdanbol (approved):

PR #1031, opened 2025-04-07
Add indicators entry point in design system

rchavan10:


libs/design-system/indicators/results-indicator/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export const greeting = 'Hello World!';
```
axdanbol, line 1: Remove

axdanbol (dismissed):

axdanbol (approved):

Issue #1021, opened 2025-03-27
Add tables


LibbyUX:
# Ready for dev
Please add our table components to our Design System Storybook:
- [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=2674-376)
- [Design System Repository](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=7791-64814&t=doMF8YMctx09RwV7-4)

## Component notes
- Columns in tables may have max widths edited in local design files to improve behavior for responsive widths.
- 3 table styles: Alternating colors, divider, and basic.
- Tables need virtual scrolling when they hit their local max height.
- Tables need overlay scrollbar and overlay fade components when they are scrollable.
- Table headers can be sort buttons or plain text without sorting functionality.

![Image](https://github.com/user-attachments/assets/fe0d7a7f-a148-4c11-918d-750ab4b88c22)

PR #1036, opened 2025-04-08
Update Table in Design System

edlu77:
- #1021


edlu77:
> Hi @edlu77 thank you for passing this over for review! We are super close! A handful of comments and questions:
> 
> 1. Is it possible to have scrolling and sorting at the same time? I was hoping it is possible, because we have some very long tables in the HRA Portal could benefit from having both. I have specified max heights for those tables, so our devs will always have that guidance for scrolling tables.
> 2. Is it possible to right align number value columns? I ask because Katy prefers right aligned number values.
> 3. I see the text hyperlink in the tables appears a darker red or maybe a different red altogether. Can we please use the specified text hyperlink color?
> 4. Would it be possible to have a toggle for vertical dividers for columns? I know it wasn't specified initially, but this could be nice for scaling purposes. Please use [this pattern](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=7784-58434&t=n59h5YUspHoeePiZ-4) for the HRA KG app as an example. By default, we don't use vertical dividers.
> 5. I am currently not using pagination on any tables in our apps. I feel like it clutters up UIs. Let me know if I need to specify that due to any of my previous comments/questions on this review.
> 
> This component is looking so clean, thanks again!

Thanks for the feedback! 

1. I can put in a toggle to enable/disable sorting. Will the max height on the tables always be 500px like in the spec? I feel like that is something that should be left up to the developer.
2. Will add this feature. Do we want to exempt this rule for the leftmost column or just apply it to all number values in the table?
3. The color for the links looks right to me. It's on-tertiary-fixed (#B20A2F) right?
4. Will add
5. I don't believe pagination will be necessary anymore with the scrolling behavior.

axdanbol (changes requested):

axdanbol (approved):

PR #1044, opened 2025-04-10
update micro tooltip to plain tooltip 

rchavan10:
update micro tooltip to plain tooltip and move to design-system/tooltips folder


CHANGELOG.md
```diff
@@ -295,7 +295,7 @@ Changelog for the Human Reference Atlas (HRA) User Interfaces
   - `footer`
   - `fullscreen`
   - `info-modal`
-  - `micro-tooltip`
+  - `plain-tooltip`
```
axdanbol, line 298: Don't change this. We'll add new entries to the changelog once we update the versions

axdanbol (changes requested):
One minor change

axdanbol (approved):

Issue #780, opened 2024-10-15
Add Rich Tooltip


LibbyUX:
# Ready for dev
Please add our new Material 3 inspired Rich Tooltip to Storybook. Please reach out to me if you have any questions!

## Component planning
- This component is needed for the HRA-KG database application. See: #738 

### Component properties
I am leaning into using different component properties in Figma to better organize the design system. New Figma component property types are used in this component:
- Booleans
- Text properties

![image](https://github.com/user-attachments/assets/0896b630-1db9-42bb-8358-e2907abbca16)

### Behavior
- Default Material 3: [Rich tooltip behavior](https://m3.material.io/components/tooltips/guidelines#2e51a05c-c989-432e-9496-0d93a507f726)
  - Users will click to open this tooltip, however, if feasible now or later, it would be nice to have the behavior of "rich tooltip appear when the page loads and a new feature is being explained".

![image](https://github.com/user-attachments/assets/80608008-3d00-4a40-886e-fedd0581e761)

## Resources
- [Design System Repo](https://www.figma.com/design/0QFfJAqnUUH1q6aucBYy3k/Knowledge-Graph-UI?node-id=115-12006&t=Wh0TmcOY9MwFxURF-4)
- [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=1782-202&t=RJJb65twBcMgkKH5-4)
- [Material 3: Tooltips Overview](https://m3.material.io/components/tooltips/overview)

saaivignesh20:
Development has been completed.

Raised PR for this issue: [https://github.com/hubmapconsortium/hra-ui/pull/1063](https://github.com/hubmapconsortium/hra-ui/pull/1063)

PR #1063, opened 2025-04-14
Add Rich Tooltip to Design System & Storybook

saaivignesh20:
This pull request resolves the GitHub issue #780

## Implementation Notes
Implemented two variants according to the requirement spec:
1. Simple Version
    - Text Content
    - Zero or one action
3. Advanced Version
    - Custom Content
    - Zero or more actions

## Screenshot
image


libs/design-system/tooltips/rich-tooltip/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './lib/rich-tooltip.directive';
+
+export * from './lib/content/rich-tooltip-content.component';
```
axdanbol, line 3: Only export the classes we want to be public

libs/design-system/tooltips/rich-tooltip/src/lib/rich-tooltip.directive.ts
```diff
@@ -0,0 +1,214 @@
...
+   * Custom content for the rich tooltip
+   * - Not required if need to use simple content variant
+   */
+  readonly customContainer = input<RichTooltipContainerComponent>(undefined, { alias: 'hraRichTooltip' });
+
+  /**
+   * Title for the rich tooltip
+   * - Not required if using custom content variant
+   */
+  readonly tagline = input<string>(undefined, { alias: 'hraRichTooltipTagline' });
```
axdanbol, line 75: Let's update the default value to the empty string `''` so it doesn't show as the string `'undefined'` when it is used in a custom content without a title element

libs/design-system/tooltips/rich-tooltip/src/lib/rich-tooltip.directive.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * Description for the rich tooltip
+   * - Not required if using the custom content variant
+   */
+  readonly description = input<string>(undefined, { alias: 'hraRichTooltipDescription' });
```
axdanbol, line 81: Let's update the default value to the empty string `''` so it doesn't show as the string `'undefined'` when it is used in a custom content without a description element

libs/design-system/tooltips/rich-tooltip/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './lib/rich-tooltip.directive';
+
```
axdanbol, line 2: Export module as well

axdanbol (changes requested):
Let's try to implement some of the accessibility requirements from https://m3.material.io/components/tooltips/accessibility

PR #1067, opened 2025-04-15
Us1 improvements

rchavan10:
Add scrolling to the table and set a max height to 26rem for predictions page.

Predictor page:
Made the gray background hug width completely and added responsiveness to cards.

### DIFF
--- a/apps/apps.humanatlas.io/src/app/app.component.scss
+++ b/apps/apps.humanatlas.io/src/app/app.component.scss
@@ -11,11 +11,3 @@
     flex-direction: column;
   }
 }
-
-@media only screen and (max-width: 944px) {
-  .content {
-    width: 100%;
-    margin: 0;
-    padding: 0;
-  }
-}
--- a/apps/apps.humanatlas.io/src/app/app.component.scss
+++ b/apps/apps.humanatlas.io/src/app/app.component.scss
@@ -11,3 +11,11 @@
     flex-direction: column;
   }
 }
+
+@media only screen and (max-width: 944px) {
+  .content {
+    width: 100%;
+    margin: 0;
+    padding: 0;
+  }
+}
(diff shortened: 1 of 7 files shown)


apps/apps.humanatlas.io/src/app/app.component.scss
```diff
@@ -11,3 +11,11 @@
     flex-direction: column;
   }
 }
+
+@media only screen and (max-width: 944px) {
```
axdanbol, line 15: You should be able to remove this. The content already spans the full with and doesn't have any margin/padding

apps/apps.humanatlas.io/src/app/pages/us1/cell-population-predictor/cell-population-predictor.component.scss
```diff
@@ -51,6 +52,7 @@
 
   .card {
     height: 15.8125rem;
+    height: auto !important;
```
axdanbol, line 55: Can you try updating the cards in the design system instead so we can avoid `!important`

apps/apps.humanatlas.io/src/app/pages/us1/cell-population-predictions/cell-population-predictions.component.scss
```diff
@@ -26,6 +26,8 @@ button {
 .table-container {
```
axdanbol, line 26: Also update the margins, etc. for the predictions page

apps/apps.humanatlas.io/src/app/pages/us1/cell-population-predictions/cell-population-predictions.component.scss
```diff
@@ -26,6 +26,8 @@ button {
...
   margin-top: 3rem;
   margin-bottom: 0.3125rem;
+  max-height: 26rem;
+  overflow-y: auto;
```
axdanbol, line 30: Only the table content should scroll. Not labels, headers, etc. Use ng-scrollbar. Check us2 tables for implementation details

axdanbol (changes requested):

apps/apps.humanatlas.io/src/app/pages/us1/cell-population-predictions/cell-population-predictions.component.scss
rchavan10: I have scheduled some time with Libby to get correct values for the margins

axdanbol (approved):

Issue #1059, opened 2025-04-14
apps.humanatlas.io - Global improvements/changes


axdanbol:
- [x] Add breadcrumbs configuration to each page (except landing page)
- [x] Update header to close menus on router navigations
- [x] Update progress bar color and add a minimum delay before the progress bar is shown
  Check if https://ngxtension.netlify.app/utilities/operators/derive-loading/ can be used for the delay

PR #1069, opened 2025-04-15
HRA - Global Improvements/Changes

saaivignesh20:
This pull request resolves the GitHub issue: #1059.

## Implementation Details
- Add breadcrumbs configuration to each page (except landing page)
- Update header to close menus on router navigations
- Update progress bar color and add a minimum delay before the progress bar is shown

### DIFF
--- a/apps/apps.humanatlas.io/src/app/utils/navigation.ts
+++ b/apps/apps.humanatlas.io/src/app/utils/navigation.ts
@@ -4,6 +4,11 @@ import { Event, EventType, Router } from '@angular/router';
 import { deriveLoading } from 'ngxtension/derive-loading';
 import { filter, switchMap } from 'rxjs';
 
+/**
+ * Determines whether the navigation is finished or not.
+ * @param event Navigation event
+ * @returns Boolean value indicating whether the navigation is finished or not.
+ */
 function isNavigationFinishEvent(event: Event): boolean {
   return [EventType.NavigationEnd, EventType.NavigationCancel, EventType.NavigationError].includes(event.type);
 }
--- a/apps/apps.humanatlas.io/src/app/utils/navigation.ts
+++ b/apps/apps.humanatlas.io/src/app/utils/navigation.ts
@@ -1,7 +1,12 @@
 import { inject, Signal } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
-import { EventType, Router } from '@angular/router';
-import { scan, distinctUntilChanged } from 'rxjs';
+import { Event, EventType, Router } from '@angular/router';
+import { deriveLoading } from 'ngxtension/derive-loading';
+import { filter, switchMap } from 'rxjs';
+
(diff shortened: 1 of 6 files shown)


apps/apps.humanatlas.io/src/app/utils/navigation.ts
```diff
@@ -1,30 +1,27 @@
 import { inject, Signal } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
-import { EventType, Router } from '@angular/router';
-import { scan, distinctUntilChanged } from 'rxjs';
+import { Event, EventType, Router } from '@angular/router';
+import { deriveLoading } from 'ngxtension/derive-loading';
+import { filter, switchMap } from 'rxjs';
+
+function isNavigationFinishEvent(event: Event): boolean {
```
axdanbol, line 7: Need to document this 😝

libs/design-system/navigation/header/src/lib/header.component.ts
```diff
@@ -125,6 +130,15 @@ export class HeaderComponent {
     });
 
     explicitEffect([this.menuOffsetPx], () => this.updateMenuPositions(), { defer: true });
+
+    this.router.events
+      .pipe(
+        takeUntilDestroyed(),
+        filter(
```
axdanbol, line 137: Use `[EventType.NavigationEnd, EventType.NavigationSkipped].includes(event.type)` instead

libs/design-system/navigation/header/src/lib/header.component.ts
```diff
@@ -125,6 +130,15 @@ export class HeaderComponent {
     });
 
     explicitEffect([this.menuOffsetPx], () => this.updateMenuPositions(), { defer: true });
+
+    this.router.events
```
axdanbol, line 134: Inject router here since it is the only place it is used. Also it should be optional so the header works in apps without the router.
```ts
inject(Router, { optional: true })?.events...
```

libs/design-system/navigation/header/src/lib/header.component.ts
```diff
@@ -114,6 +117,8 @@ export class HeaderComponent {
   private readonly mobileMenuOverlay = viewChild('mobileMenuOverlay', { read: CdkConnectedOverlay });
   /** Currently open menu or undefined */
   private readonly activeMenu = signal<Menu | 'main' | undefined>(undefined);
+  /** Reference to the application router */
+  private readonly router = inject(Router);
```
axdanbol, line 121: Inject in the constructor instead of saving to a property

axdanbol (changes requested):

apps/apps.humanatlas.io/src/app/utils/navigation.ts
saaivignesh20: Done 👍

axdanbol (approved):

PR #1071, opened 2025-04-16
added radio button example to the design system

prashantchanne12:


libs/design-system/buttons/radio-button/src/lib/global-styles.component.ts
```diff
@@ -0,0 +1,11 @@
+import { ChangeDetectionStrategy, Component, ViewEncapsulation } from '@angular/core';
+
+@Component({
```
axdanbol, line 3: Missing code docs here

axdanbol (changes requested):

axdanbol (approved):

libs/design-system/buttons/radio-button/src/lib/global-styles.component.scss
```diff
@@ -0,0 +1,12 @@
+@use '../../../../styles/utils';
+
+@include utils.global-styles() {
+  mat-radio-button {
+    --mdc-radio-selected-pressed-icon-color: var(--mat-sys-on-tertiary-fixed);
```
axdanbol, line 5: The hover/pressed color for non-checked buttons is incorrect. Please update.
Use `:checked` css selector to differentiate between the two

axdanbol (changes requested):

axdanbol (approved):

PR #1079, opened 2025-04-17
initial setup for form fields

rchavan10:
This is the initial setup for the form fields so that mat form fields follow same styles globally.


axdanbol:
Part of #1084

Issue #1042, opened 2025-04-09
Bug Report: RUI Rotation Tooltip


danqaurooni:
The RUI rotation tooltip says to hold shift to rotate in smaller degrees but the way it actually works is the opposite: the default rotation is by 1 degree and if you hold the shift key the block jumps around in larger degrees.

bherr2:
Yes, I think that's true. It was originally what was described in the tooltip, but was flipped to match how it worked with tissue block movement and the shift key. A new tooltip is in order. @LibbyUX can you get us new text to put in there?

axdanbol:
This has been fixed in develop. @bherr2 should I cherry pick the changes into main or do we wait?

Issue #986, opened 2025-03-03
Tooltips for tissue section: thickness and #sections


danqaurooni:
The thickness tooltip is missing so it is not clear what unit we are measuring thickness in. Since the blocks are in millimeters, I believe the sections should be in microns (micrometers) but @bherr2 should know better.

andreasbueckle:
I added one comment with the unit.

----------------------------------------------
Best,
Andreas

Andreas "Andi" Bueckle, Ph.D.
Research Lead, Cyberinfrastructure for Network Science Center (CNS)
Faculty Advisor, Sailing Club at Indiana University
US Sailing Certified Race Official
Room #4024
Luddy School of Informatics, Computing, and Engineering, Indiana University
700 N Woodlawn Ave (Luddy Hall)
47408 Bloomington, IN - USA

email: ***@***.******@***.***>
web: www.andreas-bueckle.com<https://www.andreas-bueckle.com/>
________________________________

PR #1089, opened 2025-04-21
Update tooltip text in tissue block settings

edlu77:
- #986 
- #1042


edlu77:
@LibbyUX I've made the changes as requested. However one issue I see with the tooltips is that there's no standard maximum width, so they have to be set manually. The benefit is you won't get the awkward wrapping issues like you see above. However it's a lot of hard coding values which might not be great either. Let me know what you think about this.

axdanbol (approved):

PR #1099, opened 2025-04-23
Added app label to the desing system

prashantchanne12:


libs/design-system/app-label/src/lib/app-label.module.ts
axdanbol, line 1: Please delete this file

libs/design-system/app-label/src/lib/app-label.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="ui-content">
```
axdanbol, line 1: Please remove the wrapping div and move any scss styles to the :host selector

axdanbol (changes requested):
Please move `app-label` folder into the `content-templates` folder.
Also replace usage of `hra-product-header` in apps.humanatlas.io with the app label

apps/apps.humanatlas.io/src/app/pages/us1/cell-population-predictor/cell-population-predictor.component.scss
```diff
@@ -19,6 +19,7 @@
   flex-wrap: wrap;
   gap: 3rem;
   margin-top: 3.5rem;
+  height: 15.8125rem;
```
axdanbol, line 22: There height changes incorrectly affects the responsiveness of the cards and page. Check the app at around 680px width and you'll see the problem

axdanbol (changes requested):

apps/apps.humanatlas.io/src/app/pages/us1/cell-population-predictor/cell-population-predictor.component.scss
axdanbol, line 1: The cards do not behave nicely when resizing the screen. They need to work from 320px & up. Please revisit the styling here. In general you should not need to set a fixed height. Auto is correct in most cases but you also try keywords such as min-content/max-content

axdanbol (changes requested):

axdanbol (approved):

PR #1100, opened 2025-04-23
RUI bug - reimplemented rotation slider using overlay in slider-box

gauri2029:

PR #1104, opened 2025-04-24
Reimplemented rotation slider using overlay for slider-box

gauri2029:
Continues work from #1100


axdanbol (changes requested):
Please implement requested changes made in #1100

Issue #1109, opened 2025-04-25
Add 500 error page


LibbyUX:
![Image](https://github.com/user-attachments/assets/8f92a419-f048-43e3-a6e3-bd7afcd96434)

# Ready for dev
Please add our new 500 error page indicator to Storybook and add this infrastructure for all HRA websites. This is the UI for all unexpected 500 error codes.
- This page recommends the user tries refreshing, otherwise, we suggest reporting an issue. Users also have the option to go back to the homepage `https://humanatlas.io/`
- The `Report issue` button opens a new tab to a new GitHub issue. It would be great if we could funnel them into a Bug issue template that is somewhat filled out already. Please let me know if I can help with this!
- Nice to have (or a scaling note for later): Ideally, it would great to have automated tracking and reporting of the error type details so we can improve our website. If we can automatically track these errors, then potentially we won't have to ask the user to submit the error.

### Two variants:
- 640px and above: Horizontal padding of 40px
- 639px to 320px: Horizontal padding of 16px

### Question on behavior & desired examples
- Is it possible for the 500 error section height to take up the whole screen? Asking because I want this design/implementation to be super straightforward to decrease cognitive load. See #909 for examples.

# Figma
- [Design System Repository](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=9589-15471&t=OyVLb2J8fENPyyeX-4)
- [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=3292-4471)

PR #1114, opened 2025-04-29
Design System - Add 500 Error Page

saaivignesh20:
This pull-request resolves the GitHub issue #1109.

## Implementation Notes
- Added 500 (internal server error) component to Design System/Error Pages
- Created a 500 route on apps.humanatlas.io


libs/design-system/error-pages/server-error-page/src/lib/server-error-page.component.ts
```diff
@@ -0,0 +1,18 @@
...
+ * - Displays a 500 error page when there is an internal server error.
+ */
+@Component({
+  selector: 'hra-server-error-page',
+  imports: [CommonModule, ButtonsModule, MatIconModule, RouterModule],
+  templateUrl: './server-error-page.component.html',
+  styleUrl: './server-error-page.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ServerErrorPageComponent {}
```
axdanbol, line 18: Add an input for the report url and update the button to go to the url

libs/design-system/error-pages/server-error-page/src/lib/server-error-page.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="title">Oops! We’re having a moment.</div>
+<div class="description">Try refreshing the page or report the issue if it continues.</div>
+<div class="actions">
+  <button mat-button hraCtaButton hraButtonVariant="primary">
+    Report issue
+    <mat-icon iconPositionEnd>arrow_forward</mat-icon>
```
axdanbol, line 6: Use `arrow_right_alt` instead

libs/design-system/error-pages/server-error-page/src/lib/server-error-page.component.html
```diff
@@ -0,0 +1,12 @@
...
+  </button>
+  <button mat-button hraCtaButton hraButtonVariant="secondary" routerLink="/">
+    Go to homepage
+    <mat-icon iconPositionEnd>arrow_forward</mat-icon>
```
axdanbol, line 10: Use `arrow_right_alt` instead

axdanbol (changes requested):

libs/design-system/error-pages/server-error-page/src/lib/server-error-page.component.ts
axdanbol: @LibbyUX Will the report button ever be hidden or is it always shown?

saaivignesh20: Thanks for the clarification @LibbyUX & @axdanbol, I will mark the `input()` required so that the report button always works when the page is shown.

Issue #1076, opened 2025-04-17
Update 3rd party angular packages


axdanbol:
The following 3rd party angular packages should be upgraded:
- [ ] ng-inline-svg-2 
- [ ] ngx-color-picker
- [ ] ngx-highlightjs

PR #1125, opened 2025-05-01
Upgrading 3rd party angular packages to latest versions.

gauri2029:
#1076  The following 3rd party angular packages are upgraded:
1. ng-inline-svg-2 to version 19.0.0
2. ngx-color-picker to version 19.0.0
3. ngx-highlightjs to version 14.0.0


libs/cde-visualization/src/index.ts
```diff
@@ -22,5 +22,10 @@ export type CdeVisualizationElementProps = InputProps<CdeVisualizationComponent>
 
 /** Custom element definition for CdeVisualizationComponent */
 export const CdeVisualizationElement = createCustomElement('cde-visualization', CdeVisualizationComponent, {
-  providers: [provideHttpClient(), provideAnimations(), importProvidersFrom(ColorPickerModule), provideDesignSystem()],
+  providers: [
+    provideHttpClient(),
+    provideAnimations(),
+    importProvidersFrom(ColorPickerDirective),
```
axdanbol, line 28: This is incorrect. Providers should only be imported from modules. Check if ngx-color-picker exports a provideColorPicker function or similar function. Otherwise remove this line

libs/design-system/color-picker/src/lib/color-picker.stories.ts
```diff
@@ -16,7 +16,7 @@ const meta: Meta = {
       providers: [],
     }),
     moduleMetadata({
```
axdanbol, line 18: Remove the entire `decorators` property. The component already imports the necessary directives, etc.

axdanbol (changes requested):

libs/cde-visualization/src/index.ts
gauri2029: I couldn't find any provide function from ngx-color-picker so i removed it,

axdanbol (approved):

PR #1128, opened 2025-05-02
Hra portal about page

edlu77:
About page for HRA Portal #1174 

Added FlexContainer component for content templates


edlu77:
@LibbyUX Thanks for the review! I've added the table of contents and navigation functionality to the page, so feel free to play around with it. I also found the emails and links for the profiles, so no need to worry about it.

I did notice one thing, Mark Halushka's link is no longer working: https://labs.pathology.jhu.edu/halushka/team/. Let me know if we need to replace it.

axdanbol (approved):

Issue #1105, opened 2025-04-24
Add yaml linting


axdanbol:
- [x] Add [eslint-plugin-yml](https://github.com/ota-meshi/eslint-plugin-yml)
- [x] Add `yaml` config in eslint.config.mjs - Should also include json schema validation
- [x] Use `yaml` config in applicable apps and libs

PR #1129, opened 2025-05-02
Tooling: Add YAML Linting

saaivignesh20:
This pull-request resolves issue #1105.

## Implementation Notes
- Added [eslint-plugin-yml](https://github.com/ota-meshi/eslint-plugin-yml)
- Added yaml config in eslint.config.mjs - Should also include json schema validation
- Use yaml config in applicable apps and libs
  - Done for `apps.humanatlas.io` and `humanatlas.io`
- Added configuration in `.vscode/settings.json` since ESLint only checks JavaScript/TypeScript files by default.


package.json
```diff
@@ -194,6 +195,7 @@
     "typescript": "5.7.3",
     "typescript-eslint": "^8.26.1",
     "webpack": "5.98.0",
+    "yaml-eslint-parser": "^1.3.0",
```
axdanbol, line 198: I think you should be able to remove this. `eslint-plugin-yml` is self contained

eslint.config.mjs
```diff
@@ -140,6 +141,7 @@ export const configs = {
       },
     },
   ],
+  yaml: [...eslintPluginYml.configs['flat/recommended'], ...jsonSchema.configs['flat/recommended']],
```
axdanbol, line 144: Also add `...eslintPluginYml.configs['flat/prettier']` to turn off rules that might interfere with prettier

.vscode/settings.json
```diff
@@ -31,5 +31,6 @@
     "lib:utils",
     "lib:webcomponents",
     "lib:state"
-  ]
+  ],
+  "eslint.validate": ["yaml", "json", "typescript"]
```
axdanbol, line 35: This is not necessary. Eslint will find the files to lint based on the config. Please remove

axdanbol (changes requested):

.vscode/settings.json
saaivignesh20: Added `javascript` after the quick discussion.

axdanbol (approved):

PR #1141, opened 2025-05-06
Material overrides updates

rchavan10:


axdanbol (changes requested):
Please replace all `../../../../../../libs/design-system/styles/vars` with `vars` in files inside the apps/ folder

axdanbol (approved):

Issue #1168, opened 2025-05-12
Data Selector Component


rchavan10:
- Currently we do not have a data selector component that switches content based on the selected value. For example, based on the release version we need to display the correct table data for summary statistics.
 - Here is the Figma reference for the 2D FTU Illustrations page: [FIgma](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=455-53902&t=SNZJOB9id3n4Ag7J-1)

PR #1170, opened 2025-05-13
Versioned Data Table component

rchavan10:
This is a new component inside content templates to toggle and select data.

It address this issue: https://github.com/hubmapconsortium/hra-ui/issues/1168


rchavan10:
This component has been renamed to Versioned Table Data component.

Issue #1182, opened 2025-05-14
Refactor table logic


axdanbol:
- [x] Add types to each column, i.e. string, number, markdown, link, etc.
       Schema:
```ts
z.object({
  id: z.string(),
  label: z.string(),
  type: z.enum([...])
})
 ```
- [x] Add `rowsUrl` input
- [x] Update `rows` input to be optional
- [x] Select data based on either `rows` or `rowsUrl` inputs
- [x] Change header to be sticky

PR #1194, opened 2025-05-15
Table Logic Refactoring

saaivignesh20:
This pull-request resolves the GitHub issue: #1182.

## Implementation Notes:
- Added types for each column, i.e. string, number, markdown, link, etc.
- Add `csvUrl` input
- Update `rows` input to be optional
- Select data based on either rows or `csvUrl` inputs

Updating the header to sticky is still pending.


libs/design-system/table/src/lib/table/table.component.html
```diff
@@ -6,35 +6,47 @@
...
     <tr mat-header-row *matHeaderRowDef="columnIds()"></tr>
     <tr mat-row [class.odd]="i % 2 === 1" *matRowDef="let row; let i = index; columns: columnIds()"></tr>
   </table>
 </ng-scrollbar>
+
+<ng-template hraTextRowElement #text let-text>
+  <span class="text">{{ text }}</span>
+</ng-template>
+
+<ng-template #numeric let-value>
```
axdanbol, line 39: Need to use the directives on numeric and markdown templates for typings

libs/design-system/table/src/lib/table/table.component.ts
```diff
@@ -1,27 +1,117 @@
...
     HraCommonModule,
-    MatTableModule,
+    MarkdownModule,
     MatSortModule,
+    MatTableModule,
     ScrollingModule,
     ScrollOverflowFadeDirective,
     TextHyperlinkDirective,
-    MarkdownModule,
+    LinkRowElementDirective,
```
axdanbol, line 113: Missing directives for numeric and markdown

libs/design-system/table/src/lib/types/page-table.schema.ts
```diff
@@ -31,25 +11,82 @@ export const TableVariantSchema = z.enum(['alternating', 'divider', 'basic']);
...
+  type: z.literal('text'),
+});
+
+/** Type for Numeric Column */
+export type NumericColumnType = z.infer<typeof NumericColumnTypeSchema>;
+
+/** Schema for Numeric Column */
+export const NumericColumnTypeSchema = z.object({
+  type: z.literal('numeric'),
+  // TODO add format
```
axdanbol, line 30: Need to complete this todo

axdanbol (changes requested):
Please update styles so that they match figma
You'll also need to update the yaml content files

Issue #1232, opened 2025-05-31
Add custom scroll


rchavan10:
Currently we are using Angular router's `withInMemoryScrolling` for page scroll when navigating to new page.
However, for some pages like _2d-ftu-illustrations_, there is unnecessary scroll to the top of page when the query params change.

Create a custom scroll service that handles scrolling and does not scroll to the top when the query params change

PR #1233, opened 2025-05-31
add custom scroll service

rchavan10:
This PR address the issue of scroll on top when query params change on some pages.
Here is the [issue](https://github.com/hubmapconsortium/hra-ui/issues/1232)


axdanbol (changes requested):
Please update the data viewer's controller to preserve url fragments, otherwise it navigates to the top of the page when changing organ/version

Issue #1118, opened 2025-04-29
Footer tweaks: Add Bluesky social icon button, reorganize icons, and reduce the size of the HRA logo


LibbyUX:
# Ready for dev
Please update our footer in Storybook and globally with the following tweaks:
- [ ] Add Bluesky icon button to open in a new tab to the CNS Bluesky account: `https://bsky.app/profile/cnscenter.bsky.social`
- [ ] Rearrange the icon buttons as specified
- [ ] Reduce the size of the HRA Brandmark to the specified

# Figma
- [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=938-464)
- [Design System Repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=6136-28173&t=NjsIfmMwrp4aqsTx-4)

![Image](https://github.com/user-attachments/assets/cb593d46-a94a-49d2-bc6b-44d418d5fc45)

PR #1312, opened 2025-06-11
Footer tweaks

edlu77:
- #1118


edlu77:
@LibbyUX Yes there were some accessibility issues on the social media icons. It's been fixed, thanks for catching it!

Issue #1309, opened 2025-06-10
Beta feedback: 9th release tasks


LibbyUX:
### apps.humanatlas.io/cde
- [ ] Visualization needs to fit in container or the container must visually indicate more is available below the cut

![Image](https://github.com/user-attachments/assets/7da7649b-4c23-44f3-b5d0-c919cafeee34)

### apps.humanatlas.io/rui
- [ ] Libby to do after 9th: Landing page/better design needed. Definitely need global with breadcrumbs.

![Image](https://github.com/user-attachments/assets/7f3d3b53-bb4a-46f7-9133-0a97e132d0c3)

PR #1321, opened 2025-06-12
Added Image component to fix issue with img in markdown

gauri2029:
- Issue #1309 
- added Image component to yaml files to display images in content pages instead of img.


axdanbol (changes requested):
The images should be moved before the markdown. Or did Libby specify that they should appear after?

PR #1327, opened 2025-06-12
Fix virtual scroll sizing issues

edlu77:


apps/ccf-eui/src/app/modules/results-browser/results-browser/results-browser.component.ts
```diff
@@ -87,6 +90,7 @@ export class ResultsBrowserComponent {
         return result;
       }
     }
+    this.viewport().checkViewportSize();
```
axdanbol, line 93: Move this into an effect in the constructor. Computed properties should be pure, i.e. have no side effects.
```ts
constructor() {
  effect(() => {
    this.items(); // Grab a dependency on items
    this.viewport().checkViewportSize();
  })
})
```

axdanbol (approved):

PR #1348, opened 2025-06-13
HRA Dashboard UI Fixes/Improvements

saaivignesh20:
### Changes
- applied fix for absolute urls in breadcrumbs
- fix line height for dashboard page paragraphs
- update breadcrumbs for apps and dashboard apps
- added tooltips for secondary navigation buttons

### DIFF
--- a/libs/dashboard/src/lib/components/dashboard-layout/dashboard-layout.component.scss
+++ b/libs/dashboard/src/lib/components/dashboard-layout/dashboard-layout.component.scss
@@ -28,7 +28,7 @@
     @include utils.use-font(body, xl);
     color: vars.$primary-fixed;
     margin-bottom: 2rem;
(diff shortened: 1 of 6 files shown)


libs/dashboard/src/lib/components/dashboard-layout/dashboard-layout.component.scss
```diff
@@ -28,7 +28,7 @@
     @include utils.use-font(body, xl);
     color: vars.$primary-fixed;
     margin-bottom: 2rem;
-    line-height: 1.3125rem;
+    // line-height: 1.3125rem;
```
axdanbol, line 31: Please remove instead of commenting out

axdanbol (changes requested):

axdanbol (approved):

PR #1411, opened 2025-06-30
Update illustration viewer

rchavan10:
This PR is to update the tooltip with the design system tooltip


axdanbol (commented):
I'm a bit confused as to what this pr is doing. It looks like you are adding a tooltip on top of the tooltip!? Could you please clarify what is supposed to happen?

axdanbol:
Stale

Issue #1387, opened 2025-06-18
Digital Object Table


edlu77:
Table to display digital object info in the Knowledge Graph

[Figma link](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=738-108108&t=dHWxO5TXbLrY7F8f-4)

![Image](https://github.com/user-attachments/assets/d158f0c9-35f4-4ce8-b6ba-649eb2f0f31a)

### Interactions
- The KG has an infinite scroll with a sticky header
   - Searching autocompletes with relevant matches
- Counter at the top right indicates how many entries are currently being displayed
- Download button at the start of each row opens a menu displaying all available download options:

![Image](https://github.com/user-attachments/assets/f6b3d75d-039b-46d1-a646-abe4bce5e262)

### Data table formatting
- All `Date last modified` dates are formatted: `YYYY-MM`
- All number values are formatted with commas: `0,000,000`
#### Data + branded iconography
- For datasets with multiple organs, please use the `All organs` icon. On hover, please use a micro tooltip that says `[num] organs` if a count is possible. If it is not possible, please use a micro tooltip that says `Multiple organs`.
- For crosswalks, we will use the digital object type connected to the ASCT+B table. For example, an FTU to ASCT+B crosswalk table will have a FTU product logo.

### Prototype video of infinite scroll + sticky header
https://github.com/user-attachments/assets/422e08b0-fbcd-476f-ad3f-5309570eb8ed

Issue #1383, opened 2025-06-18
Digital Object Filter


edlu77:
Collapsible filter drawer containing filter options for digital objects

[Figma](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=783-256684&t=iDOaBd9InPt74FGF-4)

![Image](https://github.com/user-attachments/assets/029b50a5-4b0c-4207-816d-8d46d6a84062)

- The filters are sticky on scroll
- Filters have two buttons:
   - Info icon button, opening a rich tooltip about the filter category. 
   - Category button, opening a flyout menu in which users can search and refine filters. Clicking an option in these flyouts generates a chip. 
   ![Image](https://github.com/user-attachments/assets/51cbdcea-9801-4a9e-a5aa-41eae0a05fd8)
   ![Image](https://github.com/user-attachments/assets/edaa21d8-2fcf-444b-a1bd-5c9c5eb8b511)
- The filter menu may be expanded or collapsed and on mobile a FAB appears to open filters 
   ![Image](https://github.com/user-attachments/assets/ac37fd11-90eb-4f08-befb-b589f33642eb)

PR #1430, opened 2025-07-13
Kg filter menu + table

edlu77:
- #1383 
- #1387


apps/kg-explorer/src/app/app.component.scss
```diff
@@ -1,8 +1,18 @@
 :host {
-  display: block;
+  display: grid;
```
axdanbol, line 2: This grid setup seems unnecessary and might prevent proper scrolling of the header, etc. Instead set min/max height on the content (maybe with `height: 100%`). Experiment with flex properties inside the `.content` to get the right behavior

apps/kg-explorer/src/app/components/filter-menu/filter-menu-overlay/filter-menu-overlay.component.ts
axdanbol, line 1: Parts of this component needs to be refactored into reusable design system components. I've seen the chips with autocomplete logic in multiple places now and there are subtle bugs in all of them so we need to work together to implement it as a form field compatible component (similar to the rui's slider input component)

apps/kg-explorer/src/app/components/filter-menu/filter-menu.component.ts
```diff
@@ -0,0 +1,69 @@
...
+    anatomicalStructures: new FormControl<FilterOption[] | null>(null),
+    cellTypes: new FormControl<FilterOption[] | null>(null),
+    biomarkers: new FormControl<FilterOption[] | null>(null),
+  });
+
+  readonly filterRecord = input.required<FilterOptionList>();
+  readonly drawerClosed = signal<boolean>(false);
+  readonly closeDrawer = output<boolean>();
+
+  readonly formChanges = output<FormGroup<FilterFormControls>>();
```
axdanbol, line 59: Do not emit the form itself. It is internal to the component. Instead emit the new form value

apps/kg-explorer/src/app/pages/main-page/main-page.component.ts
axdanbol, line 1: This component is far too complicated and needs to be refactored into multiple smaller pieces

libs/design-system/table/src/lib/types/page-table.schema.ts
```diff
@@ -38,20 +38,53 @@ export const MarkdownColumnTypeSchema = z.object({
...
+  type: z.literal('menu'),
+  icon: z.string(),
+  options: z.string(),
+});
+
+/** Type for MenuOptions Column */
+export type MenuOptionsType = z.infer<typeof MenuOptionsTypeSchema>;
+
+/** Schema for MenuOptions Column */
+export const MenuOptionsTypeSchema = z.object({
```
axdanbol, line 73: I'm concerned about how these options are structured when loading a csv file. Could you please explain how it will work?

axdanbol (changes requested):

PR #1445, opened 2025-07-28
Create update biomarker matrix

rchavan10:


libs/components/behavioral/src/lib/biomarker-details/biomarker-details.component.scss
```diff
@@ -33,17 +21,16 @@
     grid-template-rows: auto 3rem 5rem min-content;
   }
 
-  .table-header-group {
-    grid-area: table;
-    display: flex;
-    max-width: 100%;
-    border-bottom: 1px solid rgba(var(--light-dividers));
-  }
+  // .table-header-group {
```
axdanbol, line 24: Remove commented out code

axdanbol (changes requested):

libs/components/behavioral/src/lib/biomarker-details/biomarker-details.component.scss
rchavan10: This is old code, so keeping it for now

PR #1484, opened 2025-08-05
Fixes in FTU fullscreen mode

gauri2029:
FTU Fullscreen fixes
1. Updated labels order in mat-tabs
2. Hide fullscreen icons in sourcelist when fullscreen mode is on
3. Show mat-toggle in biomarker in fullscreen mode
4. Styles updated in case of no-data fullscreen
5. Hide source list when there is no-data


libs/ftu-ui-components/src/lib/molecules/src/lib/source-list/source-list.component.scss
```diff
@@ -34,6 +33,32 @@
...
   }
 
+  &.no-data-full-screen {
+    .source-list-header {
+      display: none;
+    }
+    ::ng-deep ftu-empty-biomarker .footer {
+      display: flex;
+      justify-content: end;
+      border-top: 0.0625rem solid var(--mat-sys-outline);
```
axdanbol, line 43: Use color variables from `vars` instead of `var(--mat-*)`

PR #1543, opened 2025-08-07
Update styles for rich tooltip

rchavan10:


libs/design-system/tooltips/rich-tooltip/src/lib/content/rich-tooltip-content.component.scss
```diff
@@ -25,5 +25,13 @@
...
   .actions {
     padding: 0 0.5rem 0.25rem 0.5rem;
+
+    button {
+      @include utils.use-font(label, medium);
+    }
   }
 }
+
+::ng-deep .mdc-button.mat-mdc-button.mat-accent {
```
axdanbol, line 35: Update the story's css rather than this

axdanbol (changes requested):

axdanbol (approved):

PR #1557, opened 2025-08-11
Only hide title in API endpoint summary, so that description can be shown.

bherr2:


saaivignesh20:
@bherr2 @LibbyUX I have made the requested changes in styles.
- The unnecessary padding is no more visible
- Optimized styles for code block
- The `<select>` element could not be styled directly, I'll keep trying to find a workaround if required in the future.

bherr2:
Looks good to me. I'll let @axdanbol review the CSS more closely.

axdanbol (approved):

Issue #1551, opened 2025-08-08
Need to add totals to table on ASCT+B table page


andreasbueckle:
@katyb requested that we add totals to the bottom of the table on https://humanatlas.io/asctb-tables#summary-statistics:

<img width="1543" height="350" alt="Image" src="https://github.com/user-attachments/assets/58e89157-36c9-446f-96ca-6569b05e3baa" />

PR #1599, opened 2025-08-21
Add total footer

prashantchanne12:
This PR adds a total footer for ASCT+B Tables.
closes https://github.com/hubmapconsortium/hra-ui/issues/1551

### New ASCT+B Table

image

### Total footer

image

### DIFF
--- a/libs/design-system/table/src/lib/table/table.component.html
+++ b/libs/design-system/table/src/lib/table/table.component.html
@@ -52,8 +52,9 @@
         }
 
         @for (column of _columns(); track column.column) {
-          @let type = getColumnType(column);
+          @let type = getColumnType(column).type;
           @let template = templates[type];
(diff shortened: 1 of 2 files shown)


libs/design-system/table/src/lib/table/table.component.html
```diff
@@ -72,13 +72,25 @@
                 *ngTemplateOutlet="template; context: { $implicit: element[column.column], element, column }"
               />
             </td>
+
+            <!-- Total Footer -->
+            <td mat-footer-cell *matFooterCellDef [attr.data-column-type]="type">
+              <span class="total-label" *ngIf="isFooterLabelColumn(column.column)">Total</span>
```
axdanbol, line 78: No `*ngIf` please, use `@if` and `@else if` blocks instead

axdanbol (changes requested):
I think we could greatly simplify the logic in the table. Let's work together when you have time to achieve this

Issue #1632, opened 2025-08-27
Update surface bright N-98 in design system to #F8FAFC


LibbyUX:
# Summary
Currently an unspecified HEX code is being used for surface bright. 

## Expected behavior
Initially this was specified as `FFFFFF`.

## Current behavior
Showing an unspecified HEX code `f5fafb`.

## Screenshot

<img width="370" height="121" alt="Image" src="https://github.com/user-attachments/assets/a934afe6-af27-487d-92ef-8f2c02a2a67a" />

# To do
- [ ] Please use `F8FAFC` for `surface bright N-98`.

Issue #1510, opened 2025-08-05
Update indicator: Error message


LibbyUX:
Libby & Daniel review this component.
### Todo
- [x] Update colors
- [x] Update to allow links in the error message
- [x] Move component into indicators module

### Figma
- Design system repo: [Error message indicator component](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=9083-27962&t=MXvo5I6anAAFLUHB-4)
- [HRA components page + instances per error type](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=1294-4977)

Issue #1498, opened 2025-08-05
Social media button


LibbyUX:
Libby & Daniel review this component.
### Todo
- [x] Add stories showcasing different variants

PR #1639, opened 2025-08-28
Design System - Bug Fixes

saaivignesh20:
This pull request resolves the following issues:
- #1498 
- #1632 
- #1510


libs/design-system/indicators/error-indicator/src/lib/error-indicator.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-icon>error</mat-icon>
+<div class="errors">
+  @for (error of errors(); track error) {
+    <div class="error">
+      {{ error }}
+    </div>
+    @if (actionLink()) {
```
axdanbol, line 7: It is still inside the for loop...

axdanbol (changes requested):

libs/design-system/indicators/error-indicator/src/lib/error-indicator.component.html
saaivignesh20: Fixed now. Thanks.

Issue #1541, opened 2025-08-07
Responsive behavior of violin graph container and overall app


LibbyUX:
# Summary
In the main view of the CDE visualization app, the violin graph pushes the distance axis downward on information dense visualizations, which hides it from view It does not have a visual indicator that scrolling is possible horizontally scrolling downwards.

## Expected behavior
In the main view of the CDE, the violin graph container ideally should show the entire visualization in the container without scrolling, even if it is small. Horizontal scrolling should not be an option. Overlay fades should be present if vertical scrolling is needed.

## Actual behavior
In the main view of the CDE, the violin graph container hides the distance axis. Users have to scroll down vertically to know that horizontal scrolling is possible.

## Todo
- [ ] Position the distance axis at the top of both the violin graph and the histogram. This was requested by Katy.
- [ ] Remove horizontal scrolling from violin graph container.
- [ ] Add a fade overlay if vertical scrolling is needed, but ideally, the violin graph is always visible in the main view.
- [ ] Adjust padding/margin in violin graph and histogram containers to account for new distance position.
- [ ] Adjust min/max widths of component containers as specified, so the app resizes responsively. Screen size notice should appear on the minimum widths of 1280 pixel widths and 810 pixel heights.

## Figma
- [Cell Distance Explorer](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=4292-289106&t=47J28jkvXCKpjBUn-4)

![Image](https://github.com/user-attachments/assets/4fca229e-49d5-4fa6-921e-c053e4323b43)

## Screenshot

<img width="1453" height="1057" alt="Image" src="https://github.com/user-attachments/assets/9797eb92-f0fb-40d1-9f9c-c1ae12fd3367" />

<img width="1480" height="989" alt="Image" src="https://github.com/user-attachments/assets/0f5e663c-6f55-4559-9ee1-8a959b5fd5aa" />

PR #1640, opened 2025-08-29
Cde layout update

edlu77:
- #1541 
- #1682

### DIFF
--- a/apps/cde-ui/src/app/pages/create-visualization-page/create-visualization-page.component.scss
+++ b/apps/cde-ui/src/app/pages/create-visualization-page/create-visualization-page.component.scss
@@ -136,6 +136,10 @@
   }
 
   .color-config {
+    mat-button-toggle-group {
+      width: fit-content;
+    }
+
     table {
       width: 66%;
 


edlu77:
@LibbyUX 

The fade overlay in the cell types table was causing problems with the scrollbar so I disabled it for now. It's somewhat complicated to fix so might need to save this for another time.

I adjusted the violin visualization so that the graph will always contain all the cell types. Since there is no more scrolling needed on this element the overlay fade won't be an issue there any longer.

Let me know if there's anything else!

libs/cde-visualization/src/lib/cde-visualization/cde-visualization.component.scss
```diff
@@ -23,22 +24,27 @@
     );
   }
 
-  .header {
-    grid-area: header;
-    border-color: vars.$outline-variant;
-    border-width: 0 0px 2px 0px;
-    border-style: solid;
-    height: 4rem;
+  // .header {
```
axdanbol, line 27: Remove commented out code

libs/cde-visualization/src/lib/components/violin/violin.component.ts
```diff
@@ -154,8 +155,14 @@ export class ViolinComponent {
...
 
+  /** If full screen mode is enabled */
+  readonly fullScreenEnabled = signal<boolean>(false);
+
+  /** Number of colors (cell types) in the visualization */
+  readonly colorCount = signal<number>(0);
+
   /** Effect for updating view data */
-  protected readonly viewDataRef = effect(() => {
+  protected readonly viewDataRef = effect(async () => {
```
axdanbol, line 165: Async functions are not allowed in `effect`

axdanbol (changes requested):

Issue #1671, opened 2025-09-08
Update text hyperlink enabled state to show underline properties


LibbyUX:
# Summary
For design system scaling and accessibility compliance across color systems, please update our text hyperlink component.

## To do
- [ ] Update properties to match Figma for all states
- [ ] Update Story in Storybook to use Nunito Sans 18px (Body XL) instead of Times New Roman

## Figma & resources
- Figma: [Design system repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=8589-27466&t=CTVGt58ZnyEaUxbh-4)
- Figma: [HRA Components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=1920-22)
- https://www.w3.org/TR/WCAG22/#use-of-color
- [Storybook](https://cdn.humanatlas.io/ui/storybook/design-system/?path=/docs/design-system-buttons-text-hyperlink--docs)

### CSS properties from Figma Dev Mode

```css
color: var(--Schemes-Tertiary-T-40, #E00D3A);

/* Hyperlink Body XL */
font-family: "Nunito Sans";
font-size: 18px;
font-style: normal;
font-weight: 400;
line-height: 27px; /* 150% */
letter-spacing: var(--Label-Letter-spacing, 0);
text-decoration-line: underline;
text-decoration-style: solid;
text-decoration-skip-ink: auto;
text-decoration-thickness: 1px; /* 5.556% */
text-underline-offset: 2px; /* 11.111% */
text-underline-position: from-font;
``` 

<img width="2307" height="982" alt="Image" src="https://github.com/user-attachments/assets/625ca57b-6b76-47cc-9bc2-8acde2d151fe" />

PR #1679, opened 2025-09-10
Update text hyperlink styles

edlu77:
- #1671


edlu77:
@LibbyUX  I've removed the underlined links for the table of contents and also the header menu. There might be other areas that need to be updated so let me know if you see any!

Issue #1706, opened 2025-09-17
Add new color variable theme to the design system


LibbyUX:
<img width="2205" height="837" alt="Image" src="https://github.com/user-attachments/assets/510ebe7d-c427-4b3d-a066-ddc92cff94ca" />

# Summary
Submitting our color variables today for CNS Light Mode. Our current colors are named HRA Light Mode. Eventually, we will add dark modes for each brand type.

## Figma
- [HRA Components: Color Variables](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=4690-3104)

<img width="765" height="1916" alt="Image" src="https://github.com/user-attachments/assets/d85ed26c-2295-4226-b891-f63c10839cab" />

PR #1843, opened 2025-10-28
Adding new CNS color variable theme to the design system

gauri2029:
- Issue linked #1706

### DIFF
--- a/libs/design-system/styles/_cns-light-theme.scss
+++ b/libs/design-system/styles/_cns-light-theme.scss
@@ -139,7 +139,3 @@ $theme: (
   ),
   typography: typo.$typography,
 );
-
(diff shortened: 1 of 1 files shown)


libs/design-system/styles/_cns-light-theme.scss
```diff
@@ -0,0 +1,145 @@
...
+$theme: (
+  color: (
+    theme-type: light,
+    primary: $_primary,
+    tertiary: $_tertiary,
+  ),
+  typography: typo.$typography,
+);
+
+@include utils.global-styles() {
```
axdanbol: Remove. The theme should not apply itself globally

axdanbol (changes requested):

Issue #1795, opened 2025-10-08
Add tune menu shell


LibbyUX:
# Summary
The tune menu is a collapsing menu with ways refine databases to particular views, sorting and grouping preferences, and filters. Initially trialed with the Knowledge Graph, we will scale this across the design system for multiple apps and brands.

# Todo
- [ ] Add tune menu

### Scaling note
Upcoming apps needing this component
- CNS website redesign
- Cell Distance Explorer
- Knowledge Graph: Replace local instance with design system instance

## Features
- Configurations within this menu alter the URL so people can share links to their favorite views, filters, and groups of datasets. Alternatively, we will link to specific configurations within the CNS mega menu's research section.
- Keyboard accessible
- Mobile accessible
- At widths of 1099px, the menu collapses automatically (like in the KG). Breakpoints at 1099px and below always receive an overlay variant with a close button when reopened.
  - Click/press outside of the mobile overlay version results in it being hidden.
- The header section is not scrollable.
- The data customizations section has overlay fade top and bottom and our overlay scrollbar when scrolling is needed.
- Tune menu content stays sticky on scroll so users can quickly access it.
- The tune menu grows in height with the page from being positioned just under the nav header at the top to the very bottom of the page.
- The tune menu may be opened or closed with a button within the component or an icon button in the global navigation menu.
- The tune menu may be oriented on either side of the layout but shows up on the left side of the database page layout by default.

## Anatomy
<img width="2624" height="2150" alt="Image" src="https://github.com/user-attachments/assets/783fefde-d74e-4a3e-a96f-14580c13122c" />

1. Collapse/close icon button: Optional. Material Symbols `arrow_menu_close` or `close` may be used to collapse the menu. Always visible if being viewed on a mobile breakpoint.
2. Header section: Optional. Used to quickly orient the user, if needed.
3. Headline: Optional headline for the database.
4. Subtext: Optional subtext for the database.
5. Overlay fade top: Not always visible and only needed for a specific scenario: User viewing menu with a short height and top overlay fade when scrolling up on content.
6. Customize section: Optional. Area to configure view of certain datasets, database view, sort by, and group by customizations.
7. Overlay scrollbar: Appears on hover when content becomes scrollable due to the height of the breakpoint.
8. Button toggle: Optional. Currently used to segment view preferences or current and former employees in the CNS website.
9. View as select: Optional. View the database as a table, gallery, or list.
[...]

Issue #1790, opened 2025-10-07
Add filter list flyout


LibbyUX:
<img width="2408" height="1080" alt="Image" src="https://github.com/user-attachments/assets/dfa08bba-2d1c-4d58-8eb3-66cb304b5ec3" />

# Summary
We need a consistent, keyboard accessible filter list flyout menu with an optional search text field with autocomplete. This is needed for scaling so we just have one design/implementation to manage across the design system. 

# Todo
- [ ] Add filter list menu flyout to the design system

## Workflow
Please complete first:
- #1787
- #1788
- #1791 

## Features
- Clicking out closes the flyout
- May be opened by an icon button or a standard button
- Search text field with autocomplete that reduces the list to match label or supporting text
- Checkbox pattern used to indicate selection
- Optional: Multi-line text
- Optional: Ripple on press turned on for list items
- Multiple items may be selected at a time 
- As users select/deselect list items, chips update under their category in the local filter menu
- Flyout must be accessible to support screen reading and keyboard navigation - Use default Material 3 options (referenced below) if supported by Angular Material
  - Enter or space to add chip/tag
  - Keyboard arrows to navigate up and down the list
- Box shadow

#### Short term
Needed for:
- CNS website in both the people database and research database
- Replace local filter menu list component in Knowledge Graph with this
  - We will replace the custom digital object instance with this pattern, too, as it could be more accessible
- RUI per #1675 as noted by @danqaurooni 
#### Long term
- Future versions may need to support multiple sections in one flyout, like on the [CellxGene Collections app](https://cellxgene.cziscience.com/collections)

<img width="1710" height="1053" alt="Image" src="https://github.com/user-attachments/assets/9d8a2db6-43ac-49d7-aa71-928b687d1176" />

### Angular Material
- [CDK: Accessibility - ListKeyManager](https://material.angular.dev/cdk/a11y/overview#listkeymanager)
- [CDK: Listbox](https://material.angular.dev/cdk/listbox/overview)
- [CDK: SelectionModel](https://material.angular.dev/components/chips/examples#chips-autocomplete)
- [Chips autocomplete](https://material.angular.dev/components/chips/examples#chips-autocomplete)
- [List: Multi-line list](https://material.angular.dev/components/list/overview#multi-line-lists)
- [List: Selection list](https://material.angular.dev/components/list/overview#multi-line-lists)

### Material 3
[...]

PR #1867, opened 2025-10-31
Filter menu shell

edlu77:
- #1795

Need this to be completed:
- #1790

### DIFF
--- a/libs/design-system/filter-menu/src/lib/filter-menu.component.html
+++ b/libs/design-system/filter-menu/src/lib/filter-menu.component.html
@@ -1,7 +1,7 @@
 @if (tagline() || description() || enableClose()) {
-  <div hraFeature="header" class="header-section">
-    <div class="drawer-header">
-      <span class="drawer-title">{{ tagline() }}</span>
+  <div hraFeature="header" class="intro section">
+    <div class="tagline">
+      <span class="title">{{ tagline() }}</span>
       <div class="filler"></div>
       @if (!isWideScreen() && enableClose()) {
         <button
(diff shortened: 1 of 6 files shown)


libs/design-system/tune-menu/src/lib/tune-menu.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  readonly filterCategories = computed(() => {
+    const result: Record<string, FormControl<Option[] | null>> = {};
+    for (const category of FILTER_CATEGORIES) {
+      result[category.id] = new FormControl<Option[]>([]);
+    }
+    return result;
+  });
+
+  readonly form = computed(() => {
+    return new FormGroup({
```
axdanbol: Form groups/controls should not be created inside of a `computed`

libs/design-system/tune-menu/src/lib/tune-menu.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  subtext?: string;
+  hideCustomizeSection?: boolean;
+  /** Categories for button toggle */
+  toggleCategories?: Option[];
+  viewAsOptions?: Option[];
+  sortByOptions: Option[];
+  groupByOptions: Option[];
+}
+
+const TOGGLE_OPTIONS = [
```
axdanbol: Dummy data should go into the spec/stories files

libs/design-system/tune-menu/src/lib/tune-menu.component.ts
```diff
@@ -0,0 +1,134 @@
...
+      }),
+      filters: new FormGroup(this.filterCategories()),
+    });
+  });
+
+  readonly formValues = computed<[string, FormControl<Option[] | null>][]>(() =>
+    FILTER_CATEGORIES.map((filter) => [filter.label, this.form().controls.filters.controls[filter.id]]),
+  );
+
+  readonly menuOptions = input<TuneMenuOptions>(DEFAULT_OPTIONS);
```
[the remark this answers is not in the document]
axdanbol: Instead of packing all options into an object let's have multiple inputs. These are the inputs you'll need:
- tagline
- description
- enableClose
- filters

Controls should be inserted with ng-content. That way we can freely customize the controls

libs/design-system/tune-menu/src/lib/tune-menu.component.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  readonly formClosed = input(false);
+
+  /** Emits when the form opening state is toggled */
+  readonly toggleForm = output();
```
axdanbol: Rename to `closeClick`

libs/design-system/tune-menu/src/lib/tune-menu.component.ts
```diff
@@ -0,0 +1,134 @@
...
+export class TuneMenuComponent {
+  readonly filterCategories = computed(() => {
+    const result: Record<string, FormControl<Option[] | null>> = {};
+    for (const category of FILTER_CATEGORIES) {
+      result[category.id] = new FormControl<Option[]>([]);
+    }
+    return result;
+  });
+
+  readonly form = computed(() => {
```
axdanbol: You don't need any form controls in this component

libs/design-system/tune-menu/src/lib/tune-menu.component.html
```diff
@@ -0,0 +1,91 @@
...
+          <hra-icon class="menu-icon" fontIcon="close" />
+        </button>
+      }
+    </div>
+    <div class="header-subtext">{{ menuOptions().subtext }}</div>
+  </div>
+}
+<ng-scrollbar hraScrollOverflowFade>
+  <form>
+    @if (!menuOptions().hideCustomizeSection) {
```
axdanbol: Replace these controls with ng-content projection. I think you can use `contentChildren('*')` to detect whether you have controls or not. Or you can add a directive that that has to be applied to projected elements. Or possibly even use https://material.angular.dev/cdk/observers/overview

libs/design-system/tune-menu/src/lib/tune-menu.component.html
```diff
@@ -0,0 +1,91 @@
+@if (!menuOptions().hideHeaderSection) {
+  <div hraFeature="filters" class="header-section">
+    <div class="drawer-header">
+      <span class="drawer-title">{{ menuOptions().headline }}</span>
+      <div class="filler"></div>
+      @if (!isWideScreen() || !menuOptions().hideCloseButton) {
+        <button
```
axdanbol: This will only be used to close the menu

axdanbol (approved):

PR #1934, opened 2025-11-14
Hra text button directive

### LINKED ISSUE #1902: Add `hraTextButton` directive
Add a directive that when used on a regular button styles it as a text hyperlink.
Usage example: https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=11918-20968&t=DYH1PYGTsd0Q3MOq-4

rchavan10:
This PR closes #1902


axdanbol (changes requested):
@rchavan10 Don't we need to override default button styles as well?
Also please add a stories file with an example.

Issue #1899, opened 2025-11-10
Improvements: Brand mark - Create and use injection token for marks


axdanbol:
Create and use an injection token for the marks just like brand logo does
- [ ] Create injection token
- [ ] Remove schema files

PR #1938, opened 2025-11-18
Brand Marks - Create and use new injection token

saaivignesh20:
This pull request resolves the issue:
- #1899

### DIFF
--- a/libs/design-system/brand/mark/src/lib/mark.component.ts
+++ b/libs/design-system/brand/mark/src/lib/mark.component.ts
@@ -19,11 +19,11 @@ export class BrandMarkComponent {
   readonly variant = input<BrandMarkVariant>('default');
 
   /** Marks from injection token */
-  readonly marks = injectBrandMarks();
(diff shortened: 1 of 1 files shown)


libs/design-system/brand/mark/src/lib/mark.component.ts
```diff
@@ -19,9 +18,12 @@ export class BrandMarkComponent {
   /** Mark variant */
   readonly variant = input<BrandMarkVariant>('default');
 
+  /** Marks from injection token */
+  readonly marks = injectBrandMarks();
```
axdanbol: Change to an input with default `input(injectBrandMarks())`

axdanbol (changes requested):

libs/design-system/brand/mark/src/lib/mark.component.ts
saaivignesh20: Done!

Issue #1954, opened 2025-11-19
Update HRA KG Explorer to work with staging/mirror sites


bherr2:
To test HRA v2.4, we create a staging API Endpoint at `https://apps.humanatlas.io/api--staging` and the data is mirrored at `https://cdn.humanatlas.io/hra-kg--staging` (the production CDN is at `https://cdn.humanatlas.io/digital-objects`).

Update the code to be able to switch to use different API endpoints and CDN URLs. It should work analogous to the EUI and RUI with environments files setting defaults that can be overwritten by web component inputs to the app.

Use the `hra-v2.4` as the base branch.

PR #2003, opened 2025-12-05
Update to set api endpoint and mirror url from HTML

edlu77:
Updates for #1954

### DIFF
--- a/apps/kg-explorer/src/app/utils/endpoints.spec.ts
+++ b/apps/kg-explorer/src/app/utils/endpoints.spec.ts
@@ -1,0 +1,60 @@
+import { TestBed } from '@angular/core/testing';
+import { Signal } from '@angular/core';
+import { HraApiConfiguration } from '@hra-api/ng-client';
+import { environment } from '../../environments/environment';
+
+import { injectRemoteApiEndpoint, setRemoteApiEndpoint, injectMirrorUrl, setMirrorUrl } from './endpoints';
+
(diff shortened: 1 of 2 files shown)


apps/kg-explorer/src/EMBEDDING.md
```diff
@@ -0,0 +1,23 @@
...
+
+The API endpoint and mirror url can be set as in this example:
+
+```html
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>HRA Knowledge Graph Explorer</title>
+    <base href="/" />
```
bherr2: @edlu77  @axdanbol  Please provide an example where the base href is on another site (ie the mirror) and the code is deployed to cdn.

apps/kg-explorer/src/EMBEDDING.md
```diff
@@ -8,7 +8,7 @@ The API endpoint and mirror url can be set as in this example:
   <head>
     <meta charset="utf-8" />
     <title>HRA Knowledge Graph Explorer</title>
-    <base href="/" />
+    <base href="https://cdn.humanatlas.io/digital-objects" />
```
bherr2: That's not what I meant @edlu77 will stop by in a minute, because that definitely doesn't work.

bherr2:
@edlu77 can this please be moved forward?

Issue #1822, opened 2025-10-16
Add stories for database app layout variant - Gallery view


LibbyUX:
# Todo
- [ ] Add stories for database app layout variant - Gallery view:
  - [ ] Gallery card variant
  - [ ] Profile card variant

## Feature summary & components needed
We are adding new dynamic ways to view database apps within the CNS website. First up is gallery view, featuring:
- #1795 
- #1791 
- #1790 
- #1753 
- #1800 
- #1805 
- Storybook: [Responsive card group](https://cdn.humanatlas.io/ui/storybook/design-system/?path=/story/design-system-cards-action-card--responsive-card-group)

This pattern must be:
- Accessible:
  - Prioritize screen reading navigation and keyboard navigation and contact me if more is needed!
- Responsive:
  - Min/max widths have been carefully designed to ensure all breakpoints work
  - Starting at 1099px: Tune menu collapses by default
  - Responsive card group from Storybook needed
- Dynamic:
  - Users can share links to their favorite filters, views, and groups
  - Search with autocomplete reduces as they type with no matches resulting in the no match indicator
    - This component is sticky and it scrolls with the page so it's always visible on all breakpoints
  - Infinite scrolling for all views
- A flexible layout for design/dev teams: We use varying cards, margin in the responsive card grid, and tune menu customizations for different database apps

### Scaling
We want to tweak the tune menu for each database app: Currently we have 3 separate front-facing database apps that need this pattern:
1. Knowledge Graph
2. CNS website: People
3. CNS website: Research

# Figma
- Design system repo - Pattern library: [Database app layout with card variants](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=12642-57693&t=Hhug8Q6yf6QcFAL9-4)
## Design file examples
- CNS design file: [Research News: Gallery view](https://www.figma.com/design/GicnRkqqY97uTPQ9WvMFXL/cns.iu.edu?node-id=488-39709&t=EyPIOnFGVleYm7xv-4)
- CNS design file: [People Current Team: Gallery view](https://www.figma.com/design/GicnRkqqY97uTPQ9WvMFXL/cns.iu.edu?node-id=424-61685&t=EyPIOnFGVleYm7xv-4)
[...]

PR #2014, opened 2025-12-11
Gallery view component

edlu77:
- #1822


LibbyUX (commented):
Thank you for making this happen @edlu77 !

I could have been clearer in my issue when I submitted #1822 - I was envisioning this story being a full page layout, including a header and footer, rather than a portion of the page. No worries if this isn't possible. If it is, should I edit #1822 to mention this?

It could be nice to see this as a full page layout, potentially, so that the background color, padding, cards behave responsively as specified in the [Figma Gallery View layout component](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=12642-57693&t=xGCXr0gCVTmeCiiy-4), with the whole layout entirely squared away. For me, this is a helpful layout template to have, so I don't have to worry about styles, margin, padding, responsive, and colors everytime I am specifying a new datapase page in Figma. Would this be a helpful component for our devs? What do you think?

### Preview
Image

### Figma
Image

Issue #2047, opened 2025-12-17
Release version filter menu sort ascending ➡️ descending


LibbyUX:
# Summary
The release version filter menu is sorting ascending (oldest first) instead of descending (latest first) for certain browsers, notably Chromium browsers, such as Google Chrome and Brave. It shows up as expected with a descending sort with Firefox.

### Chrome
<img width="1823" height="943" alt="Image" src="https://github.com/user-attachments/assets/f7a3d558-6b1a-41a3-9654-095a38ba5094" />

### Brave

<img width="2129" height="948" alt="Image" src="https://github.com/user-attachments/assets/4ba0ea88-d476-4fa1-8e78-e228bb51d45b" />

PR #2048, opened 2025-12-17
Actually sort hra versions by ids

edlu77:
- #2047


LibbyUX (approved):
Looks great on both Brave and Chrome! Thanks so much for fixing so fast!

PR #2055, opened 2026-01-02
ccf-organ-info tests

prashantchanne12:


axdanbol (changes requested):
Must use [testing-library](https://testing-library.com/docs/angular-testing-library/examples) to write the tests instead of using the `fixture` and `document`, etc.

Issue #2079, opened 2026-01-21
Research database placeholder images


LibbyUX:
<img width="1732" height="888" alt="Image" src="https://github.com/user-attachments/assets/b0e0d3ea-e327-4fd6-ba3f-418ff744b668" />

# Summary
We do not have images for many of the entries in the CNS Research database, but we are aiming to fix this by adding default images. These images are defined by category, aside from publications, which use publication type to showcase the data item on hand in gallery view. We will also use this imagery eventually in the HRA, so it needs to be added to the design system first, then configured for the CNS research data.

# Todo
- [ ] Export and add default imagery to Storybook. Note that we may want to export from Figma at 1.5x so the images look better at a higher resolution.
- [ ] Add default imagery to CNS research database items by category and if a publication, use publication type

## Figma
- Design file example: [CNS Research Database with placeholder images](https://www.figma.com/design/GicnRkqqY97uTPQ9WvMFXL/cns.iu.edu?node-id=1673-27482&t=hyqoF5mvrl4z3hj4-4)
- [Design system repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=13832-39566&t=pXQMlTyiUEphurR9-4)
- [HRA components](https://www.figma.com/design/BCEJn9KCIbBJ5MzqnojKQp/HRA-Components?node-id=5713-102)

<img width="4200" height="4479" alt="Image" src="https://github.com/user-attachments/assets/86ecd805-2cdc-4de2-9308-7a0e397899e5" />

PR #2086, opened 2026-01-28
Placeholder images

edlu77:
- #2079


bherr2:
Looks so much better with placeholder images!

apps/cns-website/src/app/utils/utils.ts
axdanbol: Please rename this file to `research-item-images.ts`

apps/cns-website/src/app/utils/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { ResearchItem } from '../schemas/research.schema';
+
+/**
+ * Gets image url and uses the appropriate placeholder if none is provided.
+ * @param item
```
axdanbol: Add documentation for the parameter and return value

axdanbol (changes requested):
A little cleanup

Issue #2075, opened 2026-01-21
Add unique item counter in filter form component


LibbyUX:
<img width="1642" height="349" alt="Image" src="https://github.com/user-attachments/assets/a5c52912-c5f4-4911-94e1-85044387bf46" />

# To do
- [x] Add toggle to switch on/off a unique item counter for the filter form category

### Behavior
- This counter displays the number of unique items within a specific category
- Designers and developers can switch the counter feature on/off, just like the leading icon button feature
- Use thousands separators for number display
### Tooltip
Add tooltip on hover of counter with preferred placement order:
1. Right middle
2. Top middle
3. Bottom middle
4. Left middle

Use tooltip text: `Number of unique items per filter category`

# Component in Figma
- [Filter form](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=8429-28659&t=lmKao1YVXiuJQgSu-4)

## Example usage for KG filter menu variant
- Figma: [HRA-KG filter menu component](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=12516-63379&t=lmKao1YVXiuJQgSu-4)
<img width="917" height="1383" alt="Image" src="https://github.com/user-attachments/assets/e407e341-163c-4921-a4cc-5480a3cb224a" />

PR #2093, opened 2026-01-29
Design System Add Unique Item Count - Filter Form

saaivignesh20:
This pull request resolves the following GitHub issue:
- #2075


libs/design-system/filter-container/src/lib/filter-container.component.scss
```diff
@@ -69,3 +71,17 @@
     margin: 0.25rem 0;
   }
 }
+
+.top-row .category-button {
```
axdanbol: Always nest styles in the top level `:host` block

axdanbol (changes requested):

Issue #2132, opened 2026-02-06
Vis app: Cell links column is cut off in cell types component


LibbyUX:
# Summary
On Brave browser at 100%, the cell links column is visually cut off, obscuring the info. I can still sort, but it is hard to read, as 1.25 numbers appear to be cut off: https://apps.humanatlas.io/cde/gallery/intestine-codex-stanford/B004_Mid_jejunum

## Todo
- [ ] Move the cells and links columns over so that the numbers are contained within the component

<img width="740" height="864" alt="Image" src="https://github.com/user-attachments/assets/e580c8a6-f518-41c5-beeb-72fad3c0e6b8" />

LibbyUX:
This might be related: https://apps.humanatlas.io/cde/gallery/bonemarrow-codex-chop/SB67_NBM34_H38

It appears that the cells aren't truncating so it pushes the cells and links columns over when there should not be a horizontal scroll.

<img width="2690" height="1424" alt="Image" src="https://github.com/user-attachments/assets/9c26a69f-1559-4621-b8f5-8e658184db33" />

PR #2137, opened 2026-02-06
Set max width on cell type column

edlu77:
Fix for #2132


LibbyUX (approved):
I checked 3 different studies and they look great. Thank you so much @edlu77 !

Issue #1892, opened 2025-11-07
Update web component with new source data property features


LibbyUX:
<img width="2252" height="717" alt="Image" src="https://github.com/user-attachments/assets/5eac2e9c-a4be-4979-aad4-141c622e2049" />

# Summary
The small web component source data table variant needs to match the new source data table in the full app layout.

# Todo
- [ ] Add new source data table properties
- [ ] Add results counters
- [ ] Update styles & responsiveness

## Features
- Updated table theme
- New UX copy for the counter - Dynamic based on user selection
- Static publications counter - Displays total publications
- New data properties from HRApop data
When row content exceeds the width/height of the responsive the viewing area:
- Sticky double header featuring the category headers + column headers
- Overlay scrollbar
  - Vertical scroll when row content exceeds the full screen window height 
  - Horizontal scroll when the table hits it's min width
- Overlay fade to indicate scrolling

## Figma
- [Small web component: Source data update 11.7.2025](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=6170-48137&t=758tGL8Y7sYudRMJ-4)

Issue #1891, opened 2025-11-07
Update full screen mode with new source data property features


LibbyUX:
<img width="2308" height="738" alt="Image" src="https://github.com/user-attachments/assets/3f64fd51-226c-4c04-9eff-58c74d5596ee" />

# Summary
The full screen mode source data table variant needs to match the new source data table in the full app layout.

# Todo
- [x] Add new source data table properties
- [x] Add results counters
- [ ] Update styles & responsiveness

## Features
- Updated table theme
- New UX copy for the counter - Dynamic based on user selection
- Static publications counter - Displays total publications
- New data properties from HRApop data
When row content exceeds the height/width of the full screen viewing area:
- Sticky double header featuring the category headers + column headers
- Overlay scrollbar
  - Vertical scroll when row content exceeds the full screen window height 
  - Horizontal scroll when the table hits it's min width
- Overlay fade to indicate scrolling

## Figma
- [Full screen mode: Source data 11.7.2025](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=6170-48136&t=758tGL8Y7sYudRMJ-4)

Issue #1887, opened 2025-11-05
More menu download options update


LibbyUX:
# Summary
Two new download types have been added to the FTU Explorer. The current menu only covers FTU illustration downloads. Users may now download source data CSV files or source data biomarker expressions JSON files. Users receive a `File downloaded` snackbar upon completion.

### FTU Explorer full app
- [ ] Update more menu's button label `Download illustration` ▶️ `Download options`
- [ ] Update current download illustration button labels:
  - [ ] `SVG` ▶️ `Illustration SVG`
  - [ ] `PNG` ▶️ `Illustration PNG`
- [ ] Add divider between the illustrations and source data downloads
- [ ] Add new download options:
  - [ ] `Source data CSV`
  - [ ] `Source data biomarker expressions JSON`
### ftu-ui-small-wc
Copy all steps above for the more menu in the small web component.
- [ ] Update more menu's button label `Download illustration` ▶️ `Download options`
- [ ] Update current download illustration button labels:
  - [ ] `SVG` ▶️ `Illustration SVG`
  - [ ] `PNG` ▶️ `Illustration PNG`
- [ ] Add divider between the illustrations and source data downloads
- [ ] Add new download options:
  - [ ] `Source data CSV`
  - [ ] `Source data biomarker expressions JSON`

# Figma
- [Design system repo](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/HRA-Design-System-Repository?node-id=12961-54433&t=UzUB4blj2YyO9YfA-4)
- [FTU Explorer canvas 11.5.2025](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=6134-174831&t=syu7V48BD5kLROQe-4)

<img width="1944" height="1390" alt="Image" src="https://github.com/user-attachments/assets/817b6809-8905-4058-a37c-f583a3c297e1" />

Issue #1882, opened 2025-11-05
Update secondary navigation download icon button interaction


LibbyUX:
# Summary
Two new download types have been added to the FTU Explorer's secondary navigation. The current button only covers FTU image downloads.

# Todo
- [ ] Update the plain tooltip from `Download illustration` to `Download options`
- [ ] Update current illustration labels:
  - [ ] `SVG` ▶️ `Illustration SVG`
  - [ ] `PNG` ▶️ `Illustration PNG`
- [ ] Add divider between the illustrations and new source data downloads
- [ ] Add new download options:
  - [ ] `Source data CSV`
  - [ ] `Source data biomarker expressions JSON`

## Figma
- [FTU Explorer canvas](https://www.figma.com/design/Jnsvqp19Flq4UFCmyHfhtz/Human-Reference-Atlas-2025?node-id=6138-179190&t=syu7V48BD5kLROQe-4)
- [Components: FTU Explorer Secondary nav]()

<img width="1340" height="709" alt="Image" src="https://github.com/user-attachments/assets/5eb0f03c-9a05-469f-a8af-fe25f26441ce" />

PR #2152, opened 2026-02-17
Update table for new source data

edlu77:
- #1879
- #1882
- #1887
- #1891
- #1892
- #2154

### DIFF
--- a/apps/kg-explorer/src/app/pages/main-page/main-page.component.scss
+++ b/apps/kg-explorer/src/app/pages/main-page/main-page.component.scss
@@ -113,6 +113,7 @@
   ::ng-deep hra-table {
     --mat-table-header-container-height: 5.375rem;
     --mat-table-row-item-container-height: 4.75rem;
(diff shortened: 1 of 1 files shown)


apps/ftu-ui-small-wc/src/styles.scss
```diff
@@ -5,3 +5,13 @@
 @include utils.global-styles() {
   @include theming.theme(light-theme.$theme);
 }
+
+mat-snack-bar-container {
```
axdanbol: Move this into the global-styles() block above

libs/ftu-ui-components/src/lib/organisms/src/lib/biomarker-table/biomarker-table.component.html
axdanbol: Did you verify that the table will never have to contain a lot of rows/columns? What is the maximum?

axdanbol (approved):

PR #2157, opened 2026-02-23
Fix profile page layout

axdanbol:


LibbyUX (approved):
Amazing, thank you @axdanbol !

PR #2158, opened 2026-02-24
refactor(app:cns-website): Refactor landing page

axdanbol:


LibbyUX (commented):
Looking good @axdanbol! Is it possible to style these tags as sentence case, just like our filters and placeholder images?

For example:
- Conference Proceeding ➡️ Conference proceeding
- Journal Article ➡️ Journal article

Image

PR #2161, opened 2026-02-25
Profile page fixes

axdanbol:


LibbyUX (commented):
Thank you @axdanbol - This is looking so good! Just a couple of questions:

### Page layout optimization
Is it possible to make the footer move down on browser resize to avoid showing a white space under the footer?

Image

### When current people do not have profile pictures:
Is it possible to hide the image or show the placeholder image? Asking because this is what I see currently:

Image

PR #2232, opened 2026-04-16
Siteimprove fixes

axdanbol:


LibbyUX (commented):
Hi @axdanbol - This looks really good. I just have one small suggestion, if possible: 
When focusing on the consent banner's `Privacy Policy` link, on desktop, the focus indicator stretches all the way to the right side buttons. Is it possible for the focus indicator border to hug the Privacy Policy link? No worries if not, just wanted to double check! It looks great otherwise!

LibbyUX:
Oh, also, I just noticed something else @axdanbol - I specified this button in the latest mobile overlay iteration, but in the context of the mega menu, this button is not shown. The CTA button to the left does the same thing in the mega menu navigation variant, so we don't need to duplicate. Is it possible to remove from the mega menu while still showing in the mobile navigation overlay? Thank you!

image

Issue #2225, opened 2026-04-13
Refactor KG Explorer to use flat files


bherr2:
To support mirroring and the upcoming time capsule, I've generated flat files for client-side filtering. Each mirror will expose a /kg/ directory from the root which will contain three files:

* [digital-objects.jsonld](https://cdn.humanatlas.io/digital-objects/kg/digital-objects.jsonld) - this is exactly the same as the old /kg/digital-objects route on HRA API.
* [asctb-terms.json](https://cdn.humanatlas.io/digital-objects/kg/asctb-terms.json) - An array of AS, CT, and B terms with ASCT+B type, iri, and label. This should be used for the associated filters
* [kg-terms-index.json](https://cdn.humanatlas.io/digital-objects/kg/kg-terms-index.json) - this exposes two big properties: `term_to_purls` and `purl_to_terms`. This will give you a lookup from digital object to terms in that DO and vice versa. This can be used for filtering based on selected terms. To save space, there is a terms and purls array which is used by the x_to_y properties as the numbers are indexes into those arrays.

The task is to replace all HRA API calls with local functions that do the same using these 3 files. For all HRA deployments and mirrors, we will use this flat file system going forward.

LibbyUX:
@edlu77 - Is this resolved? Wanted to verify. Thank you so much!

bherr2:
This is complete

Issue #2206, opened 2026-03-27
Digital object metadata page: Text field select with wrong label font size


LibbyUX:
# Summary
Our digital object metadata page is using a text field select with the wrong font size on the label: https://apps.humanatlas.io/kg-explorer/ref-organ/brain-female/latest

## Todo
- [ ] Update digital object metadata page template to use design system text field select

## Production
<img width="1382" height="420" alt="Image" src="https://github.com/user-attachments/assets/9776cf34-6587-4405-a705-e22d7d622262" />

## AtlasNG spec
- [Design System Repo | Text field](https://www.figma.com/design/gQEMLugLjweDvbsNNUVffD/AtlasNG-Design-System-Repository?node-id=6542-23071&m=dev)
<img width="1709" height="423" alt="Image" src="https://github.com/user-attachments/assets/8803bfb5-3ec1-4ad9-8f5f-cd2fbd637511" />

Issue #1967, opened 2025-11-20
kg-explorer: Increase test coverage


axdanbol:
Increase test coverage for the following regions:
- [ ] app
- [ ] app/utils

PR #2233, opened 2026-04-17
Kg explorer updates

edlu77:
- #2225
- #2239
- #2206
- #1967


apps/kg-explorer/src/app/app.component.ts
```diff
@@ -175,7 +161,15 @@ export class AppComponent extends BaseApplicationComponent {
     if (mirrorUrl) {
       setMirrorUrl(mirrorUrl);
     }
+  }
 
-    this.digitalObjects = toSignal(this.kg.digitalObjects(), { initialValue: {} });
+  private setPageTitle() {
+    this.http.get('https://cdn.humanatlas.io/digital-objects/kg/digital-objects.jsonld').subscribe((data) => {
```
bherr2: You will definitely have to refactor everywhere you directly reference https://cdn.humanatlas.io/digital-objects/ . This will be customized as an input to the app and can be changed based on the mirror it's deployed to.

apps/kg-explorer/src/app/app.routes.ts
```diff
@@ -111,11 +105,21 @@ export const appRoutes: Route[] = [
       columns: DO_COLUMNS,
     },
     resolve: {
-      data: kgResolver(),
-      asctbTermOccurrences: asctbResolver(),
-      ontologyTree: ontologyResolver(),
-      cellTypeTree: cellTypeResolver(),
-      biomarkerTree: biomarkersResolver(),
+      data: (route: ActivatedRouteSnapshot, state: RouterStateSnapshot) => {
```
axdanbol: Refactor these functions into a helper (utils/kg-resolver.ts) instead of repeating the same logic over and over

apps/kg-explorer/src/app/state/with-filters.feature.ts
```diff
@@ -0,0 +1,246 @@
...
+  digitalObjects: null,
+  releaseVersion: null,
+  organs: null,
+  anatomicalStructures: null,
+  cellTypes: null,
+  biomarkers: null,
+  searchTerm: null,
+};
+
+export function withFilters() {
```
axdanbol, line 39: This store feature should be split into 2. The first should handle the data itself, i.e. `data`, `allRows`, `versionCounts`. The second should be this feature and should only handle the filtering. This is the same structure we used for the cns website's current-team and research pages.

apps/kg-explorer/src/app/utils/utils.ts
```diff
@@ -359,3 +381,22 @@ export function sentenceCase(value: string): string {
   const processedValue = value.trim().toLowerCase();
   return processedValue.charAt(0).toUpperCase() + processedValue.slice(1);
 }
+
+export function handleValue(value: string | string[] | undefined): string[] | undefined {
```
axdanbol: Rename this to something more descriptive like `castArray` or `coerceArray`

apps/kg-explorer/src/app/services/search.service.ts
```diff
@@ -0,0 +1,91 @@
...
+import { TableRow } from '@hra-ui/design-system/table';
+import { from, Observable } from 'rxjs';
+import { TermsIndex } from '../digital-objects-metadata.schema';
+import { handleValue } from '../utils/utils';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SearchService {
+  doSearch(
```
axdanbol: Just call it `search`

apps/kg-explorer/src/app/services/search.service.ts
```diff
@@ -0,0 +1,91 @@
...
+      versions: string[];
+      ontologyTerms: string[];
+      cellTypeTerms: string[];
+      biomarkerTerms: string[];
+      digitalObjects: string[];
+      searchTerm: string | null;
+    },
+  ): Observable<string[]> {
+    const { organs, versions, ontologyTerms, cellTypeTerms, biomarkerTerms, searchTerm, digitalObjects } = options;
+
```
axdanbol, line 46: The current implementation creates a lot of intermediate arrays and iterates over the values multiple times. I would restructure something like:
```ts
search() {
  const filters = [
    this.createDigitalObjectFilter(digitalObjects),
    this.createVersionFilter(versions)
    // ...
  ];
  const result = rows.filter((item) => filters.every(fn => fn(item)));
  const ids = result.map(item => item.purl)
  return of(ids);
}

private createDigitalObjectFilter(options): (item: ItemType) => boolean {
  // Set#has is more efficient than Array#includes when there are a lot of options
  const ids = new Set(options);
  return (item) => ids.has(row.doType)
}

// Add the same for other filters
```

apps/kg-explorer/src/app/pages/metadata-page/metadata-page.component.ts
axdanbol: Use `effect` rather that observables

axdanbol: Use `effect` rather than observables

axdanbol: Put all this logic in a private helper method

axdanbol: Put all this logic into one or more private helper methods

apps/kg-explorer/src/app/pages/main-page/main-page.component.ts
```diff
@@ -150,28 +108,27 @@ export class MainPageComponent {
...
       this.populateFilterOptions();
-      this.digitalObjectSearch().subscribe((results) => {
-        this.applyMoreFilters(results);
-      });
+      this.attachDownloadOptions();
     });
 
     effect(() => {
-      this.attachDownloadOptions();
+      this.digitalObjectSearch().subscribe((results) => {
```
axdanbol: You should not subscribe inside of `effect`. Use [rxResource](https://angular.dev/api/core/rxjs-interop/rxResource#) instead

axdanbol (changes requested):

apps/kg-explorer/src/environments/environment.staging.ts
```diff
@@ -2,5 +2,5 @@
 export const environment = {
   production: true,
   remoteApiEndpoint: 'https://apps.humanatlas.io/api--staging',
-  mirrorUrl: 'https://cdn.humanatlas.io/hra-kg--staging',
```
bherr2, line 5: Staging will soon have the same flat file system, so you will want to revert that once it's up (probably friday morning).

apps/kg-explorer/src/app/utils/data/digital-object-icon-map.json
bherr2, line 1: @edlu77 rename to digital-object-icon-map.json (this can apply to more than just ASCT+B tables)

LibbyUX (approved):
Thank you so much, @edlu77 !

axdanbol (approved):

PR #2258, opened 2026-05-08
Fix issues in KG Explorer for draft versions and WPP

edlu77:


apps/kg-explorer/src/app/utils/utils.ts
```diff
@@ -230,6 +230,16 @@ export function getDocumentationUrl(doType: string): string {
...
 }
 
+/**
+ * Gets default version for a digital object
+ * @param item Digital object data item
+ * @returns default version ("draft" if available, otherwise "latest")
+ */
+export function getDefaultVersion(item: DigitalObjectInfo): string {
+  const versions = coerceArray(item.versions);
+  return versions.includes('draft') ? 'draft' : 'latest';
```
bherr2: Only pick 'draft' as the default version if it is the only version. Latest should normally go to the latest versioned DO.

axdanbol (approved):

PR #2307, opened 2026-08-05
feat(app:cns-website): team display feature requests

LibbyUX:
# Summary
- Add local component: `former-team-card`
  - Make it responsive with grayscale filter and tenure dates
- Hid redundant “Learn more” links on profiles without meaningful details
- Add desaturation effect to global design-system
- Add and modify tests


apps/cns-website/src/app/components/former-team-member-card/former-team-member-card.component.html
```diff
@@ -0,0 +1,10 @@
+<img loading="lazy" [attr.src]="pictureUrl() | assetUrl" [attr.alt]="`Profile picture of ${name()}`" />
+<div class="details">
+  <div class="name">{{ name() }}</div>
+  <div class="occupation">{{ occupation() }}</div>
+  <div class="date-ranges" role="list" aria-label="CNS tenure">
```
axdanbol: Use an ordered list instead of setting the role. Then style the list properly

apps/cns-website/src/styles.scss
```diff
@@ -9,6 +9,8 @@
 @use 'vars';
 
 @include utils.global-styles() {
+  --profile-card-image-filter: saturate(32%);
```
axdanbol: This is definitely in the wrong place

apps/cns-website/src/app/pages/current-team/state/with-people.feature.ts
```diff
@@ -28,6 +34,10 @@ export type PeopleProps = {
   endYearByPerson: Signal<Map<PeopleItem, number | null>>;
   /** Map of people to their display order */
   displayOrderByPerson: Signal<Map<PeopleItem, number>>;
+  /** Map of people to whether their profile contains meaningful additional details */
+  profileDetailsByPerson: Signal<Map<PeopleItem, boolean>>;
+  /** Map of people to their formatted continuous CNS tenure streaks */
+  tenureDateRangesByPerson: Signal<Map<PeopleItem, string[]>>;
```
axdanbol: Formatted data does not belong in here. The former team profile card should be responsible for formatting the passed data for display

apps/cns-website/src/app/pages/current-team/state/with-people.feature.ts
```diff
@@ -132,7 +146,7 @@ export function withPeople() {
       const endYearByPerson = computed(() =>
         createRolesPropertyMap(
           people(),
-          (role) => role.dateEnd && role.dateEnd.getFullYear(),
+          (role) => (isCurrentRole(role) ? null : role.dateEnd?.getFullYear()),
```
axdanbol: Pointless change that only complicates the code

apps/cns-website/src/app/pages/current-team/state/with-people.feature.ts
```diff
@@ -28,6 +34,10 @@ export type PeopleProps = {
   endYearByPerson: Signal<Map<PeopleItem, number | null>>;
   /** Map of people to their display order */
   displayOrderByPerson: Signal<Map<PeopleItem, number>>;
+  /** Map of people to whether their profile contains meaningful additional details */
+  profileDetailsByPerson: Signal<Map<PeopleItem, boolean>>;
```
axdanbol: I don't think we need to cache this

apps/cns-website/src/app/pages/current-team/current-team.component.html
```diff
@@ -80,28 +80,43 @@
...
-                    [centerContent]="true"
-                  >
-                    @if (store.team() === 'current') {
-                      <a
-                        mat-button
-                        hraHyperlink
-                        class="learn-more"
-                        [hraLink]="`/people/${person.slug}`"
-                        [attr.aria-label]="`Learn more about ${person.name}`"
+                <hra-gallery-grid
```
axdanbol, line 83: Might be time to split this into sub components to avoid deep nesting

apps/cns-website/src/app/utils/person-roles.ts
axdanbol, line 1: Code needs major improvements. It is mixing too many responsibilities. And I'm not sure if it handles some of our data assumptions correctly. Code documentation for the functions are also missing required parts like `@param` and `@returns` annotations

axdanbol (changes requested):
Needs improvements and further review, especially the logic of the tenure streak functions

apps/cns-website/src/styles.scss
LibbyUX: I intended for this to be a global style update - Does it still need edits @axdanbol?

axdanbol: Yes, it needs to be moved to the current-team scss

LibbyUX:
@axdanbol 
> Needs improvements and further review, especially the logic of the tenure streak functions

Thanks for your review yesterday! Hopefully I cleaned it up on this next pass, which is now ready for review.

apps/cns-website/src/app/components/former-team-member-card/former-team-member-card.component.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Determine whether a role begins within the month-level continuity window of a streak.
+ *
+ * @param endDate End date of the existing tenure streak
+ * @param nextStart Start date of the next role
+ * @returns Whether the role should be merged into the streak
+ */
+function canMergeWithStreak(endDate: Date, nextStart: Date): boolean {
```
axdanbol, line 57: @LibbyUX How exact do you need this to be? The current implementation will consider end date `Jan 1` and next start `Feb 28` to be a streak. Is this fine or do you want a more robust check (X number of days between)?

apps/cns-website/src/app/components/former-team-member-card/former-team-member-card.component.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Format a person's roles as continuous CNS tenure streaks, newest first.
+ * Roles belong to the same streak when their date ranges overlap or meet in the same or next calendar month.
+ * Roles with an omitted end date remain visible with an unknown end date, while null-ended current roles are excluded.
+ *
+ * @param roles Roles used to derive the tenure streaks
+ * @returns Formatted tenure ranges ordered from most to least recent
+ */
+function formatTenureDateRanges(roles: PeopleItem['roles']): string[] {
```
axdanbol, line 24: I would like this function to be split into a couple smaller parts and have some logic be placed in the `computed(() => ...)` signal. Split into `mergeStreaks` and `formatStreak`. Put the sort logic into the signal like so:
```ts
computed(() => {
  const sortedRoles = this.person().roles.filter((role) => role.dateEnd !== null).sort((a, b) => ...);
  const streaks = mergeStreaks(sortedRoles);
  return streaks.map(formatStreak);
})
```

axdanbol (changes requested):
Improve a little bit on the streak logic. Besides that it looks much better than before!

PR #2324, opened 2026-08-14
fix(lib:cde-visualization): fix histogram colors when downloading svg/png

### LINKED ISSUE #2323: Histogram plot PNG & SVG exports discard user-selected colors
# Summary
I am working on SenM graphics, and I noticed that the histogram PNG and SVG exports do not match the colors I used in the application. The UI updates with my selection, however, on export, the colors I selected are discarded.

## My visualization
Image

## Resulting files
- The violin plot correctly uses my colors while the histogram is not using the colors I chose:

Image

### PNG exports
Image
Image

axdanbol:


LibbyUX:
Thanks @axdanbol - I tested in both the wc and the app, and it works wonderfully now!

